PR #2, opened 2022-10-15
fix(integrations): make `defineLayout` auto-importable

HassanZahirnia:
This is a follow up for the https://github.com/hybridly/hybridly/commit/836d72a


innocenzi:
Good catch, thanks!

PR #11, opened 2022-10-23
docs: fix `Suggest changes to this page` link

nhedger:
The `Suggest changes to this page` link that can be found at the bottom of all documentation pages incorrectly points to the `main` branch, which leads to a 404 because the default branch is `0.x`.

Screenshot 2022-10-23 at 17 47 53

This PR: 

- [x] Updates the `editLink.pattern` option to rename the branch from `main` to `0.x`.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -50,7 +50,7 @@ export default async() => defineConfig({
 		],
 
 		editLink: {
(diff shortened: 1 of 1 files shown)


innocenzi:
I made it so that the branch name is determined programmatically. Not sure if that will work on Netlify though, let's see. Thanks for noticing the link issue!

EDIT: Turns out it doesn't work lol. I'll find another solution

PR #13, opened 2022-10-25
docs: add missing punctuation

nhedger:
Sorry for _nitpicking_.

This PR:
- [x] adds a missing question mark to the `What is Hybridly` title.


innocenzi:
It was initially on purpose but I guess it doesn't make sense to omit the question mark. Thanks!

PR #20, opened 2022-11-26
fix(routes): use directory separator to replace path

WOSHIZHAZHA120:


innocenzi:
Do you know if `base_path` returns `\\` instead of `/` as well on Windows? Or is my change enough?

WOSHIZHAZHA120:
图片
图片

windows is '\\', linux is '/'
the '\\\\' is json encoded, just replace all '\\' to '/' to resolve

the log context is json encoded, i got a wrong result before

innocenzi:
Ok, I don't have a Windows machine to test on, if you confirm the current solution works, I'll merge 👍

WOSHIZHAZHA120:
OK, all worked now
now is hope package stable and ready for production

PR #25, opened 2023-01-02
fix(types): make `Paginator` properties non-optional

HassanZahirnia:
From what I learned the Laravel's paginator always return `meta` and `links`, so I think they should be required, otherwise we have to use custom interfaces or inline types such as :
```ts
fields: {
     page: $props.posts.meta?.current_page as number,
},
```
Wasn't sure if this would be an opinionated PR, but did it anyway.

### DIFF
--- a/packages/vue/src/composables/paginator.ts
+++ b/packages/vue/src/composables/paginator.ts
@@ -8,8 +8,8 @@ declare global {
 	 */
 	interface Paginator<T = any> {
 		data: T[]
-		meta?: PaginatorMeta
-		links?: PaginatorLink[]
+		meta: PaginatorMeta
+		links: PaginatorLink[]
 	}
 
 	/**


innocenzi:
Thanks!

HassanZahirnia:
Just to leave a small note, today I encountered a situation where the `meta` was not available in the paginator data.
And this happened when I used `paginate()` in a data object, where one of the properties could be a collection of the given model's related models.
I looked up both `laravel-data` and Laravel docs, and unfortunately I couldn't find where/how this behavior would happen.

I ran into too many issues with the nested paginated collection, so I just decided to pull it out of the DTO. As there was little to no reference on how to use paginated data in a DTO.

innocenzi:
Ah, maybe I made them optional initially because of data objects actually. Have you taken a look at the [wrapping](https://spatie.be/docs/laravel-data/v2/as-a-resource/wrapping) documentation?

HassanZahirnia:
No I have not, but that seem to have different purpose and also not relevant to my issues. It's just my code logic got complex for no reason when I tried to paginate something inside a DTO. So I will just not take that route.

I think it's better to keep the merge unchanged for now since using a paginator object with `meta` and `links` included is far more common.

This could be explored more if more use cases of optional meta/links were reported.

PR #28, opened 2023-01-09
feat(testing): add `TestFileViewFinder` to support domains

nickarellano:
This pull request is to add a custom `TestFileViewFinder` in order to allow for assertHybridView() to utilize the domain syntax for views.


nickarellano:
Not thinking, I should have named this something more like `HybridlyFileViewFinder`

innocenzi:
Awesome! I think the current name is perfectly fine. The `domains` value should be configurable though, because it also is configurable in `vite.config.ts`

nickarellano:
resolvePageComponent also has domains hardcoded, am I missing where there is an override?

innocenzi:
Aah, you're right. Woops. I guess I'll refactor that later to take into account the options given to the Vite plugin.

nickarellano:
I just realized I didn't branch off of `0.x` for this PR.

PR #30, opened 2023-01-13
feat(testing): add bool assertion support to `assertHybridProperties`

HassanZahirnia:
This PR adds support to assert boolean values for assertHybridProperties helper.
```php
->assertHybridProperties([
    'foo' => true, // assert that value is true
]);
```


innocenzi:
Can you add a last test case with a `false` value?

HassanZahirnia:
Done!

innocenzi:
Thank you!

PR #32, opened 2023-01-15
feat: make Hybridly conditionable

HassanZahirnia:
This PR is to add a `when()` and `unless()` helpers, so sharing data conditionally becomes easier.

Before:
```php
if( $condition ){
    hybridly()->share('foo', 'bar');
}
```
After:
```php
hybridly()->when($condition)->share('foo', 'bar');

// or:

hybridly()->unless($condition)->share('foo', 'bar');
```

### DIFF
--- a/packages/laravel/src/Concerns/HasSharedProperties.php
+++ b/packages/laravel/src/Concerns/HasSharedProperties.php
@@ -26,12 +26,24 @@ trait HasSharedProperties
     }
 
     /**
-     * Shares data to every response if the condition is true.
+     * Only apply the chain calls if the condition is true.
      */
-    public function shareWhen(bool|\Closure $condition, string|array|Arrayable $key, mixed $value = null): static
+    public function when(bool|\Closure $condition): static
     {
         if ($condition instanceof \Closure ? $condition($this) : $condition) {
-            $this->share($key, $value);
+            return $this;
+        }
+
+        return new static();
+    }
+
+    /**
+     * Unless the condition is true, apply the chain calls.
+     */
+    public function unless(bool|\Closure $condition): static
+    {
+        if ($condition instanceof \Closure ? $condition($this) : $condition) {
+            return new static();
         }
 
         return $this;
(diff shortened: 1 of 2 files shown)


packages/laravel/src/Concerns/HasSharedProperties.php
```diff
@@ -25,6 +25,18 @@ public function share(string|array|Arrayable $key, mixed $value = null): static
         return $this;
     }
 
+    /**
+     * Shares data to every response if the condition is true.
+     */
+    public function shareWhen(bool|\Closure $condition, string|array|Arrayable $key, mixed $value = null): static
+    {
+        if ($condition) {
```
innocenzi: If `$condition` is given as a closure here, it will always pass even if the closure is supposed to return `false`.

innocenzi (changes requested):

innocenzi:
I think I would rather use https://github.com/laravel/framework/blob/9.x/src/Illuminate/Conditionable/Traits/Conditionable.php than having this. I'm not really a fan of specialized `when` methods

```php
hybridly()->when($condition)->share('foo', 'bar');
```

packages/laravel/src/Concerns/HasSharedProperties.php
HassanZahirnia: Oops, fixed it.

packages/laravel/src/Concerns/HasSharedProperties.php
```diff
@@ -25,6 +25,18 @@ public function share(string|array|Arrayable $key, mixed $value = null): static
         return $this;
     }
 
+    /**
+     * Shares data to every response if the condition is true.
+     */
+    public function shareWhen(bool|\Closure $condition, string|array|Arrayable $key, mixed $value = null): static
+    {
+        if ($condition === true || ($condition instanceof \Closure && $condition() === true)) {
```
innocenzi:
```suggestion
        if ($condition instanceof \Closure ? $condition($this) : $condition) {
```

No need to check for `true`, truthy/falsy is enough.

HassanZahirnia:
> hybridly()->when($condition)->share('foo', 'bar');

Good suggestion! Let me see if I can implement that syntax. Will let you know in a moment.

innocenzi:
The trait should be added to the `Hybridly` manager class, not some of its traits. I'll merge on Monday 👍

HassanZahirnia:
Yeah I felt off putting a trait into an other, fixed it!

PR #33, opened 2023-01-20
feat(routing): add `current`

innocenzi:
This pull request refactors the `route` function away from Ziggy's code to something I can work on. It introduces tests as well. 

More importantly, it adds a new `current` function which purpose is to check if the given named route matches the current route.

- [x] `route` tests
- [x] `route` refactor
- [x] `current` basic implementation
- [ ] `current` complex selectors (eg. `chirps.*`)
- [ ] Vue composable that returns a reactive `route` and `current` version


innocenzi:
Will add glob support for `current` and reactive versions later.

innocenzi:
@HassanZahirnia I'm warning you beforehand, `current` is very basic at the moment and has a lot of room from improvement. I didn't document it because I think it could have breaking changes soon. ^^

PR #34, opened 2023-01-24
fix(testing): use PHPUnit assertion on `assertHybridProperties`'s array syntax

HassanZahirnia:
This PR is for 2 purposes:
1. First is to rename all arrays in the `assertHybridProperties` and remove duplicate assertions. As they were overridden and basically many tests were skipped.
2. To fix the array assertion via the `assertHybridProperties` helper, as the logic used was wrong and it would fail for different type of arrays.


innocenzi:
Thanks!

PR #36, opened 2023-01-29
feat(routing): support bindings in route domain

nhedger:
This PR:

- [x] Adds support for replacing route bindings in the domain of a route.

---

### Background

Currently, the client side `route` helper only replaces existing bindings in the route's URI. 

Assuming that you define the following route in Laravel.

```php
Route::get('/', fn() => hybridly('home))
    ->domain('{subdomain}.bluebird.test')
    ->name('home')
```

And then use it in the frontend like so.

```ts
const home = route('home', { subdomain: 'demo' })
```

The resulting URL will be `https://{subdomain}.bluebird.test?subdomain=demo` while we'd expect it to be `https://demo.bluebird.test`


innocenzi:
Thank you!

nhedger:
Thank you @innocenzi !

PR #43, opened 2023-02-15
feat: support Laravel 10

saibotk:
Added the necessary dependency ranges to the composer.json and added Laravel 10 to the test workflow.

Additionally, fixed the editorconfig to use spaces in yaml (tabs would be incorrect).


saibotk:
Mhh, I'm unsure about what we want to do regarding the failing tests @innocenzi 

Do we need to have the dependencies locked in the outer scope? If yes we could adjust the workflow to allow updating those dependencies for the testing scenarios.

innocenzi:
I'm not sure yet either but I'll take a look soon, I'll need to spin up a new Laravel project so I'll want to use v10 👍

innocenzi:
Finally. Sorry if you got a lot of notifications lol. Thanks for the PR!

saibotk:
Thats no problem 😉 disabled action notifications a long time ago haha.

Thank you very much for the quick response!

PR #44, opened 2023-02-15
refactor(form): don't clear errors on reset

saibotk:
`form.reset()` is often being used to reset input elements like password inputs, when the request
failed. One example would be Laravel/Jetstream in its various password challenge forms.

With this change we are back to the behavior of inertia and would 1. make a migration easier and 2. lose nearly no DX. This is, because in the event of a successful request form errors are already not present anymore (since it was successful). The only case where `clearErrors()` now has to be called is in cases where
the UI needs to be reset, eg. when closing a modal with a form, which can still be done via the `form.clearErrors()` method.

But as already discussed on discord, I'm open for suggestions.


innocenzi:
Thanks! I'll release this change with tables in the coming days.

PR #47, opened 2023-02-17
feat(laravel): conditional return type for phpstan

saibotk:
This adds a conditional return type annotation [0] for phpstan. With this annotation, projects with phpstan enabled will not see errors anymore when using the hybridly() helper function.

Additionally we only use the phpstan specific annotation, to avoid unnecessary issues with tools like intelephense, which do not yet support conditional types.

This for example fixes:  
image

[0] https://phpstan.org/writing-php-code/phpdoc-types#conditional-return-types


innocenzi:
Thanks!

PR #51, opened 2023-02-24
fix(i18n): escape `@` for vue-i18n

nikuscs:
Currently laravel takes @ symbols on translations just normally but @ symbol at vue18n must be wrapped with brackets around it, this should bring support for it by using Str::pipe, so further symbols can also be quickly added.

@see: https://github.com/intlify/vue-i18n-next/issues/118

### DIFF
--- a/packages/laravel/src/Commands/I18nCommand.php
+++ b/packages/laravel/src/Commands/I18nCommand.php
@@ -5,7 +5,6 @@ namespace Hybridly\Commands;
 use Illuminate\Console\Command;
 use Illuminate\Support\Facades\File;
 use Illuminate\Support\Str;
-use Illuminate\Support\Stringable;
 
 /**
  * Generates a JSON file with all translations.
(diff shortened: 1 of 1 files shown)


innocenzi:
Would `replaceMatches` make the trick? Instead of `pipe` + `preg_replace`?

nikuscs:
Sure thing, im gonna take a look again asap!

nikuscs:
Just did! :) Pipe indeed was overkill.

packages/laravel/src/Commands/I18nCommand.php
```diff
@@ -100,7 +101,10 @@ protected function getTranslations(string $lang = null): array
      */
     protected function getTranslationsAsJson(string $lang = null): string
     {
-        return preg_replace('/:(\w+)/', '{${1}}', json_encode($this->getTranslations($lang)));
+        return Str::of(json_encode($this->getTranslations($lang)))
```
innocenzi:
```suggestion
        return str(json_encode($this->getTranslations($lang)))
```

PR #53, opened 2023-03-02
docs(installation): remove unnecessary import

nhedger:
This PR removes a seemingly unnecessary import from a code snippet.


innocenzi:
Good catch, thank you!

PR #59, opened 2023-03-06
docs: improve title & meta instructions

nikuscs:
Thanks to @johan and well suggested, you can define define head template directly in main.ts, this should bring some clarity on how to take advantage of this option.


innocenzi:
Thank you!

PR #61, opened 2023-03-09
docs: improve page example

salimabsi:


innocenzi:
Good catch, thanks Salim!

PR #62, opened 2023-03-10
fix(vite): skip typescript types generation on build

johanmolen:
Skip typescript types generation on build.


johanmolen:
Also fixed a small PHP deprecation:

```
Using ${var} in strings is deprecated, use {$var} instead in /vendor/hybridly/laravel/src/Commands/InstallCommand.php on line 38
```

innocenzi:
Thank you!

PR #64, opened 2023-03-16
hotfix (laravel): dialogs & database session driver

nikuscs:
When using Laravel session driver as "database" seems like the Kernel handle function for dialog request generates a new session id, this should bring a temporary hotfix to workaround the issue.

Yet is unsure if this is the best approach.

The PR ensures we store the session ID from the original request and reset it after the request has passed into the Kernel handle method.

### DIFF
--- a/packages/laravel/src/View/Factory.php
+++ b/packages/laravel/src/View/Factory.php
@@ -126,7 +126,19 @@ class Factory implements HybridResponse
         );
 
         if ($payload->dialog) {
+
+            // TODO : This is a hack to make sure the session is kept while using the database driver
+            // We should find a better way to do this in the future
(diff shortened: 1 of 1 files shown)


nikuscs:
The problem relies exactly here : https://github.com/hybridly/hybridly/blob/83cf2bc61ea18baf21cf866c881f831d56b305ec/packages/laravel/src/View/Factory.php#L162

While this doesnt seem a issue directly with hybridly, it seems like we could use another way to resolve the response for the given request.

Passing the request to handle method while using the database driver, generates a new session ID, that is then evaluated against the known sessions in database, since its its uknown, we will forced to be logged out, im not sure why the other drivers dont have this issue.

PR #65, opened 2023-03-18
fix(vite): revert skip typescript types generation on build

johanmolen:
At the moment `npm run build` does not generate `hybridly.d.ts` which can give errors if you run for example `vue-tsc` after.

The reason we disabled it on build was that `laravel-typescript-transformer` is a dev dependency.


johanmolen:
I guess this does not solve the problem entirely because `GlobalHybridlyProperties` extends `App.Data.SharedData` which is not generated. (if dev dependencies are not installed)

innocenzi:
Indeed, at the end of the day we might want `laravel-typescript-transformer` to be an actual dependency

johanmolen:
I guess so, I don't see any downsides of it.

innocenzi:
Thanks for the PR anyway!

PR #68, opened 2023-03-25
feat(dialogs): support closing without a server round-trip

innocenzi:
With this PR, you may use `router.dialog.close({ local: true })` or `const { closeLocally } = useDialog()` to close the current dialog without making an actual request to the base page.

### DIFF
--- a/docs/api/composables/use-dialog.md
+++ b/docs/api/composables/use-dialog.md
@@ -31,6 +31,12 @@ Defines whether the dialog should be shown. Generally, it is used to control the
 
 Closes the dialog, effectively assigning `false` to `show`.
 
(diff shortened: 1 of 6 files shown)


johanmolen:
Will try this on monday, thanks already!

johanmolen:
Seems like it works perfectly 👍  thanks a ton!

For the future it might be handy to close the dialog from the controller side too, what do you think? 
`back()->closeDialog()` or something?

innocenzi:
I'm not against this, but I'd need to see the use cases for this 👍

johanmolen:
Valid for now this will work more than fine 👌

PR #69, opened 2023-03-26
fix(useProperty): support `any`, arrays, `Readonly`, and Cyclic Types

djfhe:
The journey began when we encountered a "Type instantiation is excessively deep and possibly infinite.ts(2589)" bug while using `useProperty()`. Upon further investigation by @saibotk, the `any` type was identified as the root cause. We addressed the typing issue, but additional bugs emerged in the process.

To handle the growing complexity of the resulting type, we decided to extract `Path` and `PathValue` into their own dedicated package called [dot-diver 🤿](https://github.com/clickbar/dot-diver).

- [x] Resolved issues with `any` usage
- [x] Added support for arrays (e.g. `'environment.users.0.name'`)
- [x] Improved handling of cyclic dependencies between types by introducing a `depth` parameter
- [x] Tested typing and functions

The cyclic dependency aspect is rather big. In Laravel, relationships between models often introduce cyclic dependencies between the types of these models.
For example:

```typescript
type User = {
  cars: Car[]
}

type Car = {
  user: User
}
```

TypeScript will attempt to traverse these models as far as it can (currently, the default is 50 steps, i think) before throwing the same error as encountered with `any`. Our package includes a default depth parameter which sets a limit on how far TypeScript can traverse this chain. At the moment, the default value for this limit is 10, which is relatively low to protect against types becoming too large if they have numerous properties, including other objects and cyclic dependencies.

Users can still provide a longer path, but they will not receive any validation for it, and the returned value will be of an unknown type.

Additionally, we removed the generic parameter `F`. It appears that the intention was to provide the possibility to override return and parameter types. Is this functionality still needed?

### DIFF
--- a/packages/vue/build.config.ts
+++ b/packages/vue/build.config.ts
@@ -17,6 +17,7 @@ export default defineBuildConfig({
 		'hybridly',
 		'@vue/shared',
 		'axios',
+		'@clickbar/dot-diver',
 	],
 	rollup: {
 		emitCJS: true,
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -53,6 +53,7 @@
     "qs": "^6.11.1"
   },
   "devDependencies": {
+    "@clickbar/dot-diver": "^1.0.1",
     "@types/lodash": "^4.14.191",
     "@types/lodash.clonedeep": "^4.5.7",
     "@types/lodash.isequal": "^4.5.6",
--- a/packages/vue/src/composables/property.ts
+++ b/packages/vue/src/composables/property.ts
@@ -1,5 +1,7 @@
 import type { ComputedRef } from 'vue'
 import { computed, readonly } from 'vue'
+import { getByPath } from '@clickbar/dot-diver'
+import type { Path, PathValue, SearchableObject } from '@clickbar/dot-diver'
 import { state } from '../stores/state'
 import { toReactive } from '../utils'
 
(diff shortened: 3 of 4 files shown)


innocenzi:
First of all, thank you very much. Very good idea to extract this package, which I know I can use in other projects. I think you could even PR this to projects like [`type-fest`](https://github.com/sindresorhus/type-fest), [`ts-essentials`](https://github.com/krzkaczor/ts-essentials), [`utility-types`](https://github.com/piotrwitek/utility-types) or [`ts-extra`](https://github.com/sindresorhus/ts-extras). 👏 

> Additionally, we removed the generic parameter F. It appears that the intention was to provide the possibility to override return and parameter types. Is this functionality still needed?

`useProperty` accepts a first generic that would override completely its behavior: if you use `const property = useProperty<string>('some.property')`, the type of `property` would be `string` regardless of the path implementation. So if `some.property` actually exists, it would not be taken into account.

Essentially there are two generic features:

```ts
interface User {
  id: number
  full_name: string
}

interface GlobalHybridProperties {
  security: {
    user?: User
  }
  
}

// You can omit generics, in this case `security.user.id` should be type-checked, and the return type should correspond to the underlying type
useProperty('security.user.id') // number|undefined

// You can force the return type, in this case the "path" (`foo.bar`) should be a `string`, not a `PathValue<T>`
useProperty<string>('foo.bar') // string
```

I'll try this out when GitHub is up again, thank you again for the PR! ❤️

djfhe:
Yea, we removed the first generic parameter and killed the second feature by doing so. Using it disregards any type safety and there are very few cases where you really need it, since `useProperty` should be able to accept all valid paths and return the correct type.
If you still need to override any parameter/return types it is possible doing so with type assertions (which still provides at least some type safety).

However, I can understand if you want to keep this behavior, as it is more convenient than writing any explicit type assertion.

innocenzi:
In an ideal world, we wouldn't need that feature, but sometimes you have to access some property that's not typed in `GlobalHybridProperties` for instance. 

That's being said I'm fine with trying out how reliant this new implementation and wait for feedback from users to see if we have to reimplement the generic for a better escape hatch than type assertions 👍

innocenzi:
In my testing, `getByPath` works perfectly. But this PR's `useProperty` implementation doesn't seem to work, I don't have autocompletion (though I do have type-checking) on the path, and I don't have the right return type, I have an union of all possible results.

The difference seems to be that `getByPath` uses `PathEntry` and `useProperty` uses the exported `Path`.

https://user-images.githubusercontent.com/16060559/227977130-a54fe1ca-0fd7-419c-8542-97e1b259944e.mp4


Do you know why is that?

EDIT: Compiling Hybridly and using it in a project works fine, though. But I don't know why in Hybridly itself it would not work

djfhe:
Take a look at this:
https://github.com/microsoft/TypeScript/issues/26242

Typescript sometimes does not interfer your generic arguments if some of them are provided. In your example you provide Foo as a generic parameter, so the interferring of the second parameter (`P`) stops. 

It is the reason why we currently are not able to expose the generic `depth` parameter in getByPath:
https://github.com/clickbar/dot-diver/issues/1

innocenzi:
Well, today I learned. Thank you very much, I really appreciate all the time and effort put in your package and this pull request! ❤️

PR #70, opened 2023-04-09
docs(installation): improve preset usage documentation

nhedger:
This PR is an attempt at improving the documentation for using the preset.

Here's a summary of the changes:

- [x] Simplify the code snippet to only include the command for running the preset.
- [x] Add commands for `pnpm` and `yarn`
- [x] Add documentation of the flags that can be used with the preset, with links to the relevant projects.

### Simplifying the code snippet

I believe it is safe to assume that users that will run the preset to install Hybridly already know how to install Laravel, so I suggest removing the extra line that shows how to install Laravel and instead adding a link to Laravel's installation instructions.

### Add commands for `pnpm` and `yarn`

The current code snippet is updated to a code group to show the commands for the three major package managers. While the preset does not yet detect which package manager initiated the installation, I believe this is a step in the right direction.

### Add documentation of available flags

This adds a list of flags and their effect to the documentation. Users no longer have to switch to GitHub to find the information. I believe the list of flags is sufficiently small that it won't create a maintenance issue. The link to the repository is kept too.

As a bonus, I've added the missing documentation for the following flags: `--no-ide`, `--no-strict`

### DIFF
--- a/docs/guide/installation.md
+++ b/docs/guide/installation.md
@@ -6,9 +6,7 @@ You may install Hybridly in a fresh Laravel project using the preset. If you pre
 
 ## Preset
 
-The easiest way to get started with Hybridly in a [**fresh Laravel project**](https://laravel.com/docs/installation) is to use the preset.
-
-Run the following command in the root of your project.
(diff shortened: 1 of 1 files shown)


innocenzi:
Thanks! I rephrased your changes a bit and updated the preset flag table which was outdated.

nhedger:
Thank you @innocenzi !

PR #74, opened 2023-05-14
fix(forms): transform dot-notated errors to nested objects

### LINKED ISSUE #71: Form Errors Reactivity & Nested Fields Errors
### Describe the bug

There is def. something weird with the errors and im not sure yet why, but ill leave this issue to follow up

- The first i noticed while investigating some time ago, is that errors behave in a weird way in hybridly ( and a bit different from inertia ), when the backend sends the same error twice for a specific field, the errors variable is not triggered again ( and makes sense since the message is the same, and so the value never changes, so it will not trigger the watch effect. I believe this is related to original issue, reason why i wanted to mention.

- Seems like when using nested fields in forms, if we submit and the field has an error, the error is correctly set and all good here, but if you attempt to fill the said field ( so it does not contain errors ), the form.errors for that specific field seems like its never removed, while the server sends back a response without the given error shared. Its expected that also nested fields are also cleared when they are not present.

- Please do note that is also breaks the watch effect, so if you have a v-if for the given error, it will always be shown, because the variable is never removed.
 
 The only way to proper workaround this is to use the before hook to clear the given fields, but i believe this should be looked properly to have a smooth expirience.
 
 
 To note also that nested fields errors must be accessed with `form.errors['foo.nested']` instead of dot notation, ( not 100% sure but i believe it would be a nice DX to provide nested navigation for those as well.
 

### Reproduction

https://github.com/nikuscs/hybridly-demo-validation-errors

### Steps to reproduce

- Visit the https://blue-bird.test/login
- First submit without the nested field so it the error is flashed
- Then fill the contents of nested field and send again
- Errors for the given field will not be sent back by the server, but still persisted in errors
- Watch console for proper errors.

### System information

```bash
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M2
    Memory: 201.41 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.8.0 - ~/Library/Caches/fnm_multishells/68334_1681237407108/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.2.0 - ~/Library/Caches/fnm_multishells/68334_1681237407108/bin/npm
  Browsers:
    Brave Browser: 112.1.50.114
    Chrome: 111.0.5563.146
    Firefox: 110.0
    Safari: 16.4
```

### Used package manager

pnpm

### Logs

_No response_

### Validations

nhedger:
This PR is an attempt at fixing #71.

---

This PR updates the `setErrors`, `clearError` and `clearErrors` functions exposed by the `useForm` composable.

### `setErrors`

`setErrors` now expands dot-notated error keys and places the error message in an object structure that follows the structure of the form's fields. This change does not impact simple forms that do not use nesting, and they'll continue working as is.

#### Current behaviour
Assuming that all fields of the following form are required.

```ts
const form = useForm({
    url: '/test',
    method: 'POST',
    fields: {
        name: '',
        phone: {
            work: ''
        }
    }
})
```

The errors returned by Laravel would be:

```json
{
  "name": "The name field is required.",
  "phone.work": "The phone.work field is required."
}
```

and the error object tracked by Hybridly would be:

```json
{
  "name": "The name field is required.",
  "phone.work": "The phone.work field is required."
}
```

#### New behaviour

`setErrors` expands the dot-notated keys and sets the error message in the correct place.

The errors returned by Laravel are the same:

```json
{
  "name": "The name field is required.",
  "phone.work": "The phone.work field is required."
}
```

But the errors tracked by Hybridly become:

```json
{
  "name": "The name field is required.",
  "phone": {
    "work": "The phone.work field is required."
  }
}
```

### `clearError`

`clearError` now accepts a dot-notated path to the error to clear.

```ts
const form = useForm({
    url: '/test',
    method: 'POST',
    fields: {
        name: '',
        phone: {
            work: ''
        }
    }
})

form.clearError('phone.work')
```

When clearing an error, if the parent object is left empty (no more errors), the parent object is also cleared. In this example, the `phone` object will be cleared if the `work` error has been cleared.

---

Two new `setValueAtPath` and `unsetPropertyAtPath` utilities have been created and tests have been created for them.

### DIFF
--- a/packages/vue/src/composables/form.ts
+++ b/packages/vue/src/composables/form.ts
@@ -2,7 +2,7 @@ import isEqual from 'lodash.isequal'
 import type { Progress, UrlResolvable, HybridRequestOptions } from '@hybridly/core'
 import type { DeepReadonly } from 'vue'
 import { computed, reactive, ref, toRaw, watch } from 'vue'
-import { clone } from '@hybridly/utils'
+import { clone, setValueAtPath, unsetPropertyAtPath } from '@hybridly/utils'
 import { router } from '@hybridly/core'
 import { state } from '../stores/state'
 
@@ -177,7 +177,7 @@ export function useForm<T extends Fields = Fields>(options: FormOptions<T>) {
 	/**
 	 * Clears all errors.
 	 */
-	function clearErrors(...keys: (keyof T)[]) {
+	function clearErrors(...keys: string[]) {
 		if (keys.length === 0) {
 			keys = Object.keys(fields)
 		}
@@ -201,18 +201,8 @@ export function useForm<T extends Fields = Fields>(options: FormOptions<T>) {
 	/**
 	 * Clears the given field's error.
 	 */
-	function clearError(key: keyof T) {
-		delete errors.value[key]
-	}
-
-	function setNestedValue(obj: any, path: string, value: any) {
-		const segments = path.split('.')
-		let nestedObject = obj
-		for (let i = 0; i < segments.length - 1; i++) {
-			const key = segments[i]
-			nestedObject = nestedObject[key] = nestedObject[key] || {}
-		}
-		nestedObject[segments[segments.length - 1]] = value
+	function clearError(key: string) {
+		unsetPropertyAtPath(errors.value, key)
 	}
 
 	/**
@@ -220,14 +210,8 @@ export function useForm<T extends Fields = Fields>(options: FormOptions<T>) {
 	 */
 	function setErrors(incoming: Record<string, string>) {
 		clearErrors()
-		Object.entries(incoming).forEach(([key, value]) => {
-			// if the key is a dot-notated path, we expand it to set the error
-			// on the correct field
-			if (key.includes('.')) {
-				setNestedValue(errors.value, key, value)
-			} else {
-				errors.value[key] = value
-			}
+		Object.entries(incoming).forEach(([path, value]) => {
+			setValueAtPath(errors.value, path, value)
 		})
 	}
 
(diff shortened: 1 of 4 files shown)


packages/vue/src/composables/form.ts
```diff
@@ -177,7 +177,7 @@ export function useForm<T extends Fields = Fields>(options: FormOptions<T>) {
 	/**
 	 * Clears all errors.
 	 */
-	function clearErrors(...keys: (keyof T)[]) {
+	function clearErrors(...keys: string[]) {
```
innocenzi: Do you think we can have auto-completion there using `@clickbar/dot-diver` like [`useProperty`](https://github.com/hybridly/hybridly/blob/0.x/packages/vue/src/composables/property.ts)?

innocenzi (commented):
That's really cool, thank you!

packages/vue/src/composables/form.ts
nhedger: I'll have a look, but I agree it would be neat.

nhedger: This is proving harder than expected. 

Wondering if we could merge this as-is and open another PR for the auto-completion.

innocenzi: I'll take a look myself tomorrow and see if we can merge then. 👍

nhedger: Sounds good, thanks!

innocenzi:
I fixed all the typings. Thank you for this PR!

nhedger:
Awesome, thank you @innocenzi !

PR #75, opened 2023-05-15
[BUGFIX] Laravel data lazy class not found issue

shuvroroy:
Screenshot 2023-05-15 at 8 32 22 PM

### DIFF
--- a/packages/laravel/composer.json
+++ b/packages/laravel/composer.json
@@ -23,9 +23,10 @@
     ],
     "require": {
         "php": "~8.1.0 || ~8.2.0",
+        "illuminate/contracts": "^9.0|^10.0",
+        "spatie/laravel-data": "^3.5",
         "spatie/laravel-package-tools": "^1.15.0",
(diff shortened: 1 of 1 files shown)


innocenzi:
Thanks! The proper fix is to check if the class exists before using it. It's purposefully not included by default because it's supposed to be optional. It's rather unusual to install Hybridly without the preset so I didn't catch this before. Thank you!

shuvroroy:
> Thanks! The proper fix is to check if the class exists before using it. It's purposefully not included by default because it's supposed to be optional. It's rather unusual to install Hybridly without the preset so I didn't catch this before. Thank you!

Should I update the ServiceProvider Lazy macro?

innocenzi:
You shouldn't have anything to do if you installed `laravel-data`, and the fix was included in https://github.com/hybridly/hybridly/commit/8d5b205c110ea6d17a69d7a2eda95f70da098d44 (not released yet though)

PR #78, opened 2023-06-11
fix(architecture): properly compute relative file paths on Windows

salimabsi:
On Windows, `vue` files are not properly parsed due to the hardcoded forward slash `/` in the vue files finder, a proper fix is to use the `DIRECTORY_SEPARATOR` instead.

### DIFF
--- a/packages/laravel/src/Support/VueViewFinder.php
+++ b/packages/laravel/src/Support/VueViewFinder.php
@@ -169,7 +169,7 @@ final class VueViewFinder
                 if (str($path)->endsWith($this->extensions)) {
                     $files[] = [
                         'namespace' => $namespace,
-                        'path' => str_replace(base_path(DIRECTORY_SEPARATOR), '', $path),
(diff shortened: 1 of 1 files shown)


packages/laravel/src/Support/VueViewFinder.php
```diff
@@ -169,7 +169,7 @@ protected function findVueFiles(string $directory, string $baseDirectory, string
                 if (str($path)->endsWith($this->extensions)) {
                     $files[] = [
                         'namespace' => $namespace,
-                        'path' => str_replace(base_path('/'), '', $path),
+                        'path' => str_replace(base_path(DIRECTORY_SEPARATOR), '', $path),
```
innocenzi:
```suggestion
                        'path' => str_replace(base_path(\DIRECTORY_SEPARATOR), '', $path),
```

innocenzi:
Thanks!

PR #85, opened 2023-07-15
feat(routing): add pattern matching for `current` helper

saibotk:
Also adds the functionality to get the current route name, when calling the function without parameters.

The implementation heavily inspired by ziggy and is mainly built around regex matching and will transform the route definitions to regexes which are then used to check against the current URL.

Additionally some test cases are provided, but requires more realworld testing.

Also note that the mode parameter was removed from the current function as it should not be needed anymore.

---

Another thought also came to my mind, after seeing this implementation, which i would like to discuss and maybe tackle in another PR if feasible:

Since we are much more tightly coupled in Hybridly than Ziggy is with Inertia, we could also remove this frontend guessing and regex matching and if possible replace it with a server-side implementation, which provides the current route name + parameters. Would this be possible or reasonable? 
I think it could greatly improve the experience and remove nearly all of the code in this PR :)

Still i think we can evaluate this option later and for now merge this. 
I'm happy either way.

Let me know your thoughts!


innocenzi:
I'm thinking I'd rather have these helpers on the `router` interface 👀 

```ts
router.matches('index') // true
router.current() // index
```

What do you think? It wouldn't be directly compatible with Ziggy but that's not really something I'm interested in, and a match-and-replace would be enough anyway.

Your second suggestion was actually the first thing I tried, but I ended up not being comfortable transferring the route names over HTTP headers. Happy to revisit in the future though.

innocenzi:
In addition to the change suggested above, I added a `useRoute` composable and documented all utils.

Can you tell me if you can see anything wrong with this? Otherwise I'm ready to merge 👍

saibotk:
Amazing 🤩 

That actually makes me tick another todo of my list. If you wouldn't have implemented the composable that would have been the next thing i wanted to add. 🙏 

Also im very fine with the changes as discussed in Discord i am open for both override or a distinct behavior of the current and match functions.

I think this is ready to go! Thanks for shaping the final form!

nhedger:
Thanks for your work on this @saibotk and @innocenzi.

PR #86, opened 2023-07-26
docs: fix view path typo in installation guide

godismyjudge95:
Attempting to convert an Inertia.js app to Hybridly and ran into an issue with the provided example of where to create Vue pages.


innocenzi:
Thanks and sorry about that!

PR #87, opened 2023-07-27
docs: change `GlobalProperties` to `SharedData`

godismyjudge95:
Not sure if this is an accurate change but it's confusing to see SharedData in some places and GlobalProperties in others.  Might be left over from a renaming?


nhedger:
Appears to be called `SharedData` in the preset.

https://github.com/hybridly/preset/blob/main/templates/base/app/Data/SharedData.php.

My guess is it should be `SharedData` everywhere.

innocenzi:
Hey, good catch! Like Nicolas said, it should actually be `SharedData` everywhyre. I made the change, thanks for the PR!

PR #90, opened 2023-08-11
Normalize URLs for Windows

Repo-SK:
Normalize URLs so hybridly imports them as expected on Windows.


innocenzi:
I think this should be done on the PHP side? We use `DIRECTORY_SEPARATOR` here: https://github.com/hybridly/hybridly/blob/0.x/packages/laravel/src/Support/VueViewFinder.php#L164

Also I'm curious because Hybridly was reportedly working on Windows (one of the devs in my team use it) 🤔

innocenzi:
Tracking in https://github.com/hybridly/hybridly/pull/91

PR #91, opened 2023-08-15
fix: normalize vue file paths

Repo-SK:
Normalize vue file paths so it works on windows as expected


packages/laravel/src/Support/VueViewFinder.php
```diff
@@ -176,7 +176,9 @@ protected function findVueFiles(string $directory, string $baseDirectory, string
             }
         }
 
-        return $files;
+        return collect($files)
```
innocenzi: Shouldn't this only replace slashes in `path`? Also, why not replace `\DIRECTORY_SEPARATOR` directly instead of post-processing the `$files` array?

godismyjudge95: I think the question would be does `scandir` work with `/` on Windows.  Cause if not then you do have to post process the files array so the directory scanning works properly.

iyashpal: Yes, `scandir` works with `/` on windows. This snippet worked for me on windows:
```php
$files = [];

foreach (scandir($directory) as $file) {
    if (\in_array($file, ['.', '..'], true)) {
        continue;
    }

    $path = $directory . '/' . $file;

    if (is_dir($path)) {
        $files = array_merge($files, $this->findVueFiles($path, $baseDirectory, $namespace));
    } else {
        if (str($path)->endsWith($this->extensions)) {
            $files[] = [
                'namespace' => $namespace,
                'path' => str_replace(base_path(\DIRECTORY_SEPARATOR), '', $path),
                'identifier' => $this->getIdentifier($path, $baseDirectory, $namespace),
            ];
        }
    }
}

return $files;
```

innocenzi:
Thank you @Repo-SK and @iyashpal. It's funny that a feature made to handle platform-agnostic situations ended up being responsible for a platform-specific bug. ^-^

PR #95, opened 2023-08-31
fix(vite): rebuild types on middleware changes

jeffreyvanhees:
There is a typo in run.ts :)


innocenzi:
Damn, thank you, good catch!

PR #103, opened 2023-10-05
docs: createHead example correction

iyashpal:
Given `createHead` example with `titleTemplate` is not working. And, throwing this type error:

> Vue: Argument of type `{ titleTemplate: (title: any) => string;}`  is not assignable to parameter of type `CreateHeadOptions` Object literal may only specify known properties, and `titleTemplate` does not exist in type `CreateHeadOptions` 

So, I updated the example with different approach, and it worked.


innocenzi:
Good catch, thank you! This is a bit inconvenient, I'll file an issue on `@unhead/vue`. FYI, this was valid syntax before `@vueuse/head` was deprecated in favor of unhead.

PR #106, opened 2023-10-15
fix(routing): avoid matching unrelated routes using `current`

nhedger:
TBD

### DIFF
--- a/packages/core/src/routing/current.ts
+++ b/packages/core/src/routing/current.ts
@@ -23,9 +23,11 @@ function getCurrentUrl() {
  */
 export function currentRouteMatches<T extends RouteName>(name: T, parameters?: RouteParameters<T>): boolean {
 	// We escape all dots and replace all stars with a regex that matches any character sequence to build the regex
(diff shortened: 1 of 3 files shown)


saibotk (approved):
Looks good to me, seems like I missed this when implementing this.

Reference:
Ziggy also does this:
https://github.com/tighten/ziggy/blob/c53c891ede223c53046c6a8f7635a86eb017fe2a/src/js/Router.js#L122

innocenzi:
What happens if you change the order of `baz.nested` and `baz` in the fixtures? Does the test still work?

nhedger:
If by changing the order you mean the following, then yes, the test still passes.

```json
"baz.nested": {
    "domain": null,
    "method": ["GET", "HEAD"],
    "uri": "baz/nested",
    "name": "baz.nested",
    "bindings": [],
    "wheres": []
},
"baz": {
    "domain": null,
    "method": ["GET", "HEAD"],
    "uri": "baz",
    "name": "baz",
    "bindings": [],
    "wheres": []
},
```

innocenzi:
Awesome, thank you!

PR #107, opened 2023-10-20
fix: put testing helpers in autoload

johanmolen:
Load the testing helper in the `autoload` instead of `autoload-dev`.

https://getcomposer.org/doc/04-schema.md#autoload-dev
`autoload-dev` is root-only.


innocenzi (approved):

innocenzi:
Such a weird oversight on my part. Thank you Johan!

PR #108, opened 2023-11-13
fix(tables): add `IsHideable` and `EvaluatesClosures` concerns to `Group` refiner

kakasplatt:
…n() and Call to undefined method Hybridly\Refining\Group::evaluate()

### DIFF
--- a/packages/laravel/src/Refining/Group.php
+++ b/packages/laravel/src/Refining/Group.php
@@ -9,6 +9,8 @@ use Illuminate\Contracts\Database\Eloquent\Builder;
 class Group implements Refiner
 {
     use Components\Concerns\Configurable;
+    use Components\Concerns\EvaluatesClosures;
+    use Components\Concerns\IsHideable;
 
     public function __construct(
         protected array $refiners = [],


innocenzi:
I'm curious what do you need these for? `hidden` will not be inherited to the grouped refiners with just this PR, and `evaluate` is not used either.

Do you have a snippet that would help me understand your use case?

kakasplatt:
The use case is as follows, inside class extends Table:

```
/**
 * Método que define os refinadores.
 *
 * @return array Refinadores.
 */
protected function defineRefiners(): array
{
    return [
        Sort::make('id'),
        Sort::make('name'),
        RecordsPerPage::make(recordsPerPage: $this->recordsPerPage)->default('5'),
        Group::make([
            Filter::make('description', 'search')->hidden()->loose(),
            Filter::make('id', 'search')->hidden(),
            Filter::make('name', 'search')->loose()
        ])->booleanMode('or'),
    ];
}
```

I tried to use the Group method observing commit 88cd1b9 to group a single search field (input in front-end) that searches for the columns of a model. Without this “fix”, when using the code above it throws the error Call to undefined method Hybridly\Refining\Group::isHidden() and Call to undefined method Hybridly\Refining\Group::evaluate().

Unless what I want to do is not being done correctly. I apologize for my ignorance.

innocenzi:
That sounds like a bug on Hybridly's side! I'll investigate and make a release soon.

kakasplatt:
Sure. If it's a bug related to hybridly itself, the problem is located in the `getRefiners()` method at `packages/laravel/src/Tables/Concerns/RefinesAndPaginateRecords.php:28`:

```php
public function getRefiners(): Collection
{
    return $this->cachedRefinements ??= collect($this->defineRefiners())->filter(static fn (Refiner $refiner): bool => !$refiner->isHidden());
}
```

Changing it to:

```php
public function getRefiners(): Collection
{
    if (!isset($this->cachedRefinements)) {
        $is_not_hidden = static fn ($refiner): bool => in_array(IsHideable::class, class_uses_recursive($refiner)) && !$refiner->isHidden();

        $this->cachedRefinements = collect($this->defineRefiners())->filter(static function ($refiner) use ($is_not_hidden): bool {
            if ($refiner instanceof Group) {
                $refiner->refiners(collect($refiner->getRefiners())->filter($is_not_hidden)->all());

                return true;
            }

            return $is_not_hidden($refiner);
        });
    }

    return $this->cachedRefinements;
}
```

Solves the problem without implementing the traits `Components\Concerns\EvaluatesClosures` and `Components\Concerns\IsHideable` in `packages/laravel/src/Refining/Group.php`.

What do you think?

innocenzi:
That makes sense, this wasn't tested with the tables, which has specific refining logic. Thank you!

PR #109, opened 2023-11-15
fix(tables): improves the logic of obtaining refiners to allow the us e of filter grouping in RefinesAndPaginateRecords trait and remove unnecessary traits from Group class (rollback the pull request #108).

kakasplatt:
This commit is directly related to pull request #108.
When using a Group of Filters, the hidden() method stops working completely. With these modifications, it starts working at full capacity again. I had misinterpreted the functionality of hidden()…


innocenzi:
It seems your latest commit goes way beyond the scope of the original PR 

I won't add "multi sorting", at least not with this implementation (the original "single sort" implementation can be updated to support multiple sort keys instead)

Also, I wouldn't be able to merge this anyway because you updated the code style of some unrelated parts and didn't follow the conventions already in place, sorry.

Can you explain how you use `Group` in your app? So I can understand why you want these changes

innocenzi:
I'll close this PR for now, feel free to open another with a reduced scope. I think I'll end up tackling this issue myself at some point anyway.

PR #113, opened 2024-01-23
feat(tables): support custom model resolution logic for inline actions

owenconti:
Similar to the case described here: https://github.com/hybridly/hybridly/pull/112

Same as the other PR, I am happy to update this PR with a fix if you can point me in the direction you want it fixed.

---

Use case:

We allow users to view their soft deleted records. From this page, they have the ability to restore these records.

## Issue

The model look up logic in `InvokedActionController` can't handle soft deleted records. The issue is on this line:

```
$model = $table->getModelClass();
$record = $model::findOrFail($data->recordId);
```

## Solving the issue

Ideally there would be a way to define how the action's model is queried for/found. This would allow us to apply a scope during the look up.

### DIFF
--- a/packages/laravel/src/Tables/Actions/Concerns/ResolvesModel.php
+++ b/packages/laravel/src/Tables/Actions/Concerns/ResolvesModel.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Hybridly\Tables\Actions\Concerns;
+
+use Hybridly\Tables\Actions\DataTransferObjects\InlineActionData;
+use Illuminate\Database\Eloquent\Model;
+
+trait ResolvesModel
+{
+    protected \Closure $resolveModelUsing;
+
+    /**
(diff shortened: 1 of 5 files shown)


innocenzi:
> Ideally there would be a way to define how the action's model is queried for/found. This would allow us to apply a scope during the look up.

I like that. My initial reaction was to simply handle the "soft deleted" scenario instead of simply calling `findOrFail`, but your suggestion is more flexible. 

It could be a `resolveModelUsing` method on the Action instance, which can be called in the `InvokedActionController` endpoint since the action is already resolved at that point.

Would you mind PRing that?

owenconti:
@innocenzi Alright, I've pushed up my first attempt for it. Let me know what you think. Not sure what is going on with the Netlify build, too.

innocenzi:
Thanks, Owen. Looks good to me.

Maybe in the future I'll change this implementation to be more like the bulk action one where you can get the query object without executing it first, but for now this will do 👍

PR #116, opened 2024-02-02
test: update setup, cover `useForm`'s `isDirty`

owenconti:
This PR introduces a new test class for testing the `form` composable, specifically just the `isDirty` ref for now. 

While trying to add this test class, I ran into a few errors with how the tests have been configured. The root of the problem lies within the `happy-dom` package, issue here: https://github.com/capricorn86/happy-dom/issues/1180. Essentially, `happy-dom` does not implement the Request/Response classes like the browser does. `msw` relies on the browser's implementation when mocking requests.

To get around this issue for now, I've downgraded `vitest` to 0.34.6. However, this only fixed the issue locally for me. When running in CI, I was getting the [same error](https://github.com/hybridly/hybridly/actions/runs/7762676494/job/21173522365#step:10:1927). It turns out that in addition to downgrading `vitest`, I also had to update Node to v18 from v16. Hopefully that isn't a big deal. 

I've gone through the diff and left comments on the individual pieces with reasons for each change.


packages/core/test/router/__snapshots__/history.test.ts.snap
```diff
@@ -1,13 +1,13 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`serializer > serializes and unserializes complex objects 1`] = `"{"json":{"set":[1,2,3,"string",{"foo":"bar"}],"foo":"bar","nested":{"foo":"bar","set":["a","b",1,2,3,"c"]},"nestedTwice":{"foo":{"set":[1,2,3,"soleil"]}},"empty":null,"map":[["a","test1"],["b","test2"]]},"meta":{"values":{"set":["set"],"nested.set":["set"],"nestedTwice.foo.set":["set"],"empty":["undefined"],"map":["map"]}}}"`;
+exports[`serializer > serializes and unserializes complex objects 1`] = `"{\\"json\\":{\\"set\\":[1,2,3,\\"string\\",{\\"foo\\":\\"bar\\"}],\\"foo\\":\\"bar\\",\\"nested\\":{\\"foo\\":\\"bar\\",\\"set\\":[\\"a\\",\\"b\\",1,2,3,\\"c\\"]},\\"nestedTwice\\":{\\"foo\\":{\\"set\\":[1,2,3,\\"soleil\\"]}},\\"empty\\":null,\\"map\\":[[\\"a\\",\\"test1\\"],[\\"b\\",\\"test2\\"]]},\\"meta\\":{\\"values\\":{\\"set\\":[\\"set\\"],\\"nested.set\\":[\\"set\\"],\\"nestedTwice.foo.set\\":[\\"set\\"],\\"empty\\":[\\"undefined\\"],\\"map\\":[\\"map\\"]}}}"`;
```
owenconti, line 3: I am not sure about this change. I do not understand what I would have changed that would have caused the snapshot to fail.

After more debugging, I believe this is due to the change in the `vitest` version.

packages/core/test/router/history.test.ts
```diff
@@ -2,8 +2,8 @@ import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from 'vites
 import { createSerializer, setHistoryState } from '../../src/router/history'
 import { fakeRouterContext, makeRouterContextOptions, returnsArgs } from '../utils'
 
-beforeEach(() => {
-	fakeRouterContext()
+beforeEach(async() => {
+	await fakeRouterContext()
```
owenconti, line 6: Ensuring we wait for the context to be set up before starting the tests

packages/core/test/router/router.test.ts
```diff
@@ -1,29 +1,29 @@
...
-	fakeRouterContext()
-	vi.stubGlobal('console', {
-		warn: vi.fn(),
-		error: vi.fn(),
-		log: vi.fn(),
-	})
+import { fakeRouterContext, fakePayload, mockSuccessfulUrl } from '../utils'
+import { createServer } from '../server'
+
+const server = createServer()
```
owenconti: I've removed the `setup.ts` file that was used to set up the server in all of the tests. I was running into some sort of async state issue when running the entire test suite, which now has multiple calls to the mock server.

This fix of creating the server per test file seemed to fix the issue, and should also mean we don't create the server for tests that won't use it.

packages/core/test/router/router.test.ts
```diff
@@ -1,29 +1,29 @@
-import { beforeEach, it, vi } from 'vitest'
+import { beforeEach, it } from 'vitest'
 import { performHybridNavigation, router } from '../../src/router/router'
 import { getRouterContext, registerHook } from '../../src'
-import { fakeRouterContext, fakePayload, mockUrl } from '../utils'
-
-beforeEach(() => {
-	fakeRouterContext()
-	vi.stubGlobal('console', {
-		warn: vi.fn(),
```
owenconti, line 9: These mocked functions were hiding an error that was being thrown during the request to the mocked server that said that there was no handler defined for the given request.

To define the handlers, you must either pass them in when calling `setupServer`, or you can call `resetHandlers()` and pass them in to an existing server.

packages/core/test/utils.ts
```diff
@@ -27,8 +26,7 @@ export function makeRouterContextOptions(options: PartialDeep<RouterContextOptio
 		payload: fakePayload(),
 		adapter: {
 			resolveComponent: noop,
-			swapDialog: noop,
-			swapView: noop,
+			onViewSwap: noop,
```
owenconti, line 30: `swapDialog` and `swapView` no longer exist, replaced with `onViewSwap`

packages/core/test/utils.ts
```diff
@@ -38,14 +36,40 @@ export async function fakeRouterContext(options: PartialDeep<RouterContextOption
...
+/** Mocks a request using MSW. */
+export function mockUrl(options: Partial<MockOptions> = {}): RequestHandler {
 	const resolved: MockOptions = defu(options, {
 		status: 200,
 		headers: { [HYBRIDLY_HEADER]: 'true' },
 		json: fakePayload(),
 	})
 
-	http.get(url, () => {
+	return http.all('*', () => {
```
owenconti: Figured it'd be easier to mock all outgoing requests. Definitely okay to revert this though and make it more explicit.

owenconti:
So I'm running into an issue running the tests locally. Only on the __first run__ after installing the dependencies, the tests fail. If I immediately re-run the tests, then they pass.

The issue is that the `router` import from `@hybridly/core` in `form.ts` returns undefined on the first run. As far as I can tell, `vite` is supposed to use the alias and find `@hybridly/core` via the relative path.

packages/core/test/router/__snapshots__/history.test.ts.snap
innocenzi: Yes, I remember their snapshot format changed in a recent version 👍

packages/core/test/utils.ts
innocenzi: I prefer explicitness, but it's fine to let this global mock for now

owenconti: No worries, I can change it!

packages/core/test/setup.ts
```diff
@@ -1,9 +1,6 @@
-import { beforeAll, afterAll, afterEach } from 'vitest'
-import { server } from './server'
+import { afterAll, vi } from 'vitest'
 
-afterAll(() => server.close())
-afterEach(() => server.resetHandlers())
-
-beforeAll(() => {
-	server.listen({ onUnhandledRequest: 'error' })
+afterAll(() => {
```
owenconti: I need to remove these. They are available via vitest config

.github/workflows/tests.yml
```diff
@@ -65,6 +65,9 @@ jobs:
       - name: Install Node dependencies
         run: pnpm install
 
+      - name: Vitest version
+        run: cat node_modules/vitest/package.json
```
owenconti: Need to remove this

packages/core/src/index.ts
```diff
@@ -1,5 +1,5 @@
-export { createRouter, router } from './router'
-export type { Router, HybridRequestOptions, NavigationResponse, HybridPayload, ResolveComponent, Method, Progress } from './router'
+export { createRouter, router } from './router/router'
```
owenconti, line 1: This fixed an issue I was having where the tests would fail on the first run after installing the dependencies.

innocenzi: That's interesting, do you know why this specifically caused the issue?

innocenzi:
That's awesome. Thanks a lot Owen!

packages/core/src/index.ts
owenconti: Not sure. I had found some Google results about needing to `import * as x` but that only applies to default exports, could be a phantom fix.

PR #117, opened 2024-02-09
fix(tables): instanciate data instances using models instead of their serialized representation

owenconti:
This PR fixes two issues that were caused by the same source:

1. If you were using a casted property in a `Data` class, such as a `CarbonInterface`, an exception would be thrown when trying to render the `Data` class in a `Table`.
2. The `authorization` array on all Table records was always empty.

Both of these problems happened because the `Data` class was being created with a serialized JSON array of the model instead of the model itself.

The casted properties did not work because the casting of the model to JSON would transform the casted property to its primitive value (string), therefore the type would no longer match what was defined in the `Data` class.

The `authorization` array was always empty because the `Data` class was not being given a model as its source, therefore it could not call the authorization gates correctly.

### DIFF
--- a/docs/guide/tables.md
+++ b/docs/guide/tables.md
@@ -440,6 +440,28 @@ const $props = defineProps<{
 }>()
 ```
 
+You also have the ability to customize the `Data` record's resolution logic. For example, if you need to exclude authorizations (on the root record or nested records) when rendering a table:
+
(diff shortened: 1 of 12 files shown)


innocenzi:
Awesome. Thanks once again Owen!

PR #119, opened 2024-03-06
ci: add provenance to npm builds

saibotk:
This commit adds provenance for all published packages. See the NPM documentation [0].

Provenance will allow people to verify that the hybridly packages were actually built on GH Actions and with the content of the corresponding commit. This will help with supply chain security.

For this to work, the `id-token` permission was added only where necessary.

[0]: https://docs.npmjs.com/generating-provenance-statements

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,10 @@ on:
     tags:
       - "v*"
 
+permissions:
+  contents: write
+  id-token: write
+
 jobs:
   release:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -5,6 +5,10 @@ on:
     tags:
       - "v*"
 
+permissions:
+  contents: write
```
Disservin, line 9: Why write ? Neither https://docs.npmjs.com/generating-provenance-statements nor https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions, mention `contents` having write permissions, only id-token.

saibotk: This is because auf the releaser script so kt can create a release

innocenzi:
Thanks! Still confused about that permission, will take a look.

PR #121, opened 2024-03-13
feat(architecture): use kebab for view identifiers

Tofandel:
In 0.7 the components are now in lowercase

https://github.com/hybridly/hybridly/commit/fa34630e194b15a240799e26d0371e0b6b306fe3

Appart from being quite the breaking change, this now makes hard to read component names

Eg before

`Auth.ForgotPassword`
`AppPagination`

Now

`auth.forgotpassword`
`apppagination`

Proposed in PR

`auth.forgot-password`
`app-pagination`

I also put backward compat in place to ease the transition

### DIFF
--- a/packages/laravel/src/Support/VueViewFinder.php
+++ b/packages/laravel/src/Support/VueViewFinder.php
@@ -243,13 +243,17 @@ final class VueViewFinder
      */
     protected function getIdentifier(string $path, string $baseDirectory, string $namespace): string
     {
-        return str($path)
-            ->after($baseDirectory)
-            ->ltrim('/\\')
-            ->replace(['/', '\\'], '.')
-            ->replace($this->extensions, '')
+        return str(
+            str($path)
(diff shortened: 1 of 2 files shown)


innocenzi:
Personally I would prefer this as well, but that was the case at the beginning and was modified later on, I think it caused a bug somewhere but it's not specified in the commit: https://github.com/Tofandel/hybridly/commit/4fc8e199baa11ff719c6bf9c6b0807191dbb1d3d

I'll have to look it up in the discord before merging this

Tofandel:
I can tell you what that is, which I fixed in the PR with explode

With the implementation you showed

`Auth.ForgotPassword` would be `auth.-forgot-password`

And here is the relevant discussion on discord

image

The backward compat with deprecation should clear up any confusion

innocenzi (approved):

innocenzi:
Thank you!

PR #122, opened 2024-03-13
feat(architecture): allow excluding directories when loading views

Tofandel:


packages/laravel/config/hybridly.php
```diff
@@ -42,6 +42,8 @@
         'application_main' => Architecture::APPLICATION_MAIN,
         'root_view' => Architecture::ROOT_VIEW,
         'extensions' => ['vue', 'tsx'],
+        // Don't load vue files as views if they are under the following directories
+        'excluded_views_directories' => ['Components'],
```
innocenzi: Would you mind changing this to lowercase to keep consistency with the existing conventions?

innocenzi (changes requested):

packages/laravel/config/hybridly.php
Tofandel: In this case we don't need anything by default, because any directory with the name of the component or layout directory will already be excluded, this is just a default so it can handle both uppercase and lowercase

PR #124, opened 2024-03-13
Allow the vite plugin to be loaded from any directory

Tofandel:
This allows to have the vite.config file and the package.json in any directory by passing the path of laravel to the hybdridly vite plugin as a second parameter

This provides numerous advantages:

1. The structure is less rigid and allows for a cleaner root (easily 10 files less: tailwind config, postcss config, vite config, package.json, tsconfig...) (We can now recommend to keep them under resources)
2. Huge performance improvement of vite overall because the root of vite is not directly linked to laravel and all of it's files, instead the watchers are added outside of the root path to the needed files like the hybridly config, routes and app/Http (I previously had a memory leak and a huge memory footprint because of this)
3. Smaller overhead of the hybridly virtual config because the components don't need to all be prepended with `resources`

This is completely optional of course and doesn't contain breaking changes

Vite can also be kept in the root and we can have the basePath set to architecture.root_path by default which would still provide the mentionned performance boost

Example usage to have vite's route in resources without moving the vite config ```hybridly({}, {basePath: 'resources'})```

Example usage to have vite's route in resources with the vite config in resources ```hybridly({}, {laravelPath: '..'})```


packages/vite/src/config/load.ts
```diff
@@ -1,11 +1,12 @@
-import { execSync } from 'node:child_process'
+import { spawnSync } from 'node:child_process'
+import path from 'node:path'
 import type { DynamicConfiguration } from '@hybridly/core'
 
-export async function loadConfiguration(): Promise<DynamicConfiguration> {
+export async function loadConfiguration(laravelPath, basePath): Promise<DynamicConfiguration> {
```
innocenzi: Missing types here.

packages/vite/src/index.ts
```diff
@@ -14,12 +15,16 @@ import { warnOnLocalBuilds } from './local-build'
 
 type Options = ViteOptions | ((config: DynamicConfiguration) => (ViteOptions | Promise<ViteOptions>))
 
-export default async function plugin(options: Options = {}) {
-	const config = await loadConfiguration()
-	const resolvedOptions = typeof options === 'function'
+export default async function plugin(options: Options = {}, laravelPath = process.cwd(), basePath = process.cwd()) {
```
innocenzi: I'd rather have the two new arguments in an object instead, that can avoid a breaking change later.

packages/laravel/src/Commands/PrintConfigurationCommand.php
```diff
@@ -10,7 +10,7 @@
 
 class PrintConfigurationCommand extends Command
 {
-    protected $signature = 'hybridly:config {--pretty}';
+    protected $signature = 'hybridly:config {--pretty} {buildPath?}';
```
innocenzi: `buildPath` here is inconsistent with `basePath` used in the other changes. I'd rather have `basePath` I think.

innocenzi (changes requested):

innocenzi:
Tests for the changes in the view finder would be great as well 👍

packages/laravel/src/Commands/PrintConfigurationCommand.php
Tofandel: Indeed thanks for the review

Tofandel:
That's weird tests are failing on phpcs but
image

innocenzi:
Hey @Tofandel, I'm not against this feature (I think it's a good idea), but I've refactored a bit that part of the architecture implementation in #128, so I suggest we revisit it when this is merged. This is going to be for v0.8 since it has minor breaking changes.

Tofandel:
From what I see it would only cause minor merge conflicts, nothing that affects it, do you mean the refactor has breaking changes or my PR? (because normally the PR doesn't have any)

But if you are refactoring it, then maybe you could look into loading modules only when needed (this is what I did in this PR with the callbacks) but it could be done higher up for instance

We only need the view finder in the `hybridly:config` command and it's a bit of a performance issue to have the provider run the module loader which finds all vue files on every request when they are not needed anywhere except that command

Maybe something in the command that calls the registerModules of the provider and don't call it otherwise or a module manager

innocenzi:
I definitely agree with that and I should implement that back in my PR 👍

innocenzi:
I followed your callback approach, I think it was good. Not sure about implementing the "base directory" right now, but still open to it in the future

Tofandel:
We can also do the base directory transform directly in the vite plugin if you think it's more appropriate

PR #127, opened 2024-03-14
cd(provenance): use `package.json` setting instead

saibotk:
PNPM sadly does not seem to properly apply the ENV variable to the packages.

So we added the publish config to every package.json. This should actually work now, see [0] for an example.

Additionally, this removes the env var and the public flag, because this one was also moved to the config.

[0]: https://github.com/clickbar/eslint-config/blob/d88e413c610e478a48ccd52f5a4a35cfa4865760/packages/vue/package.json

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,10 +36,9 @@ jobs:
         run: pnpm run build
 
       - name: Publish to NPM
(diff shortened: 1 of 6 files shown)


innocenzi:
Thanks!

PR #128, opened 2024-03-18
refactor(architecture): improve architecture implementation

innocenzi:
This pull request improves the architecture of the architecture system (lol).

- The `getViewFinder` method on the `Hybridly` singleton no longer exists, as all `ComponentsResolver` (ex `VueViewFinder`) calls are forwarded instead of wrapped.

- The identifiers used to load views in controllers may now be generated in a custom way, using `Hybridly#setIdentifierGenerator`.

- Views, layouts, components and TypeScript directories are now lazy-loaded. This improves the performance of all Hybridly apps, since they are only needed at build-time. This has been contributed by @Tofandel in #124.


packages/laravel/src/Architecture/LazyComponentsResolver.php
```diff
@@ -1,10 +1,10 @@
...
 
-namespace Hybridly\Support;
+namespace Hybridly\Architecture;
 
 use Hybridly\Support\Configuration\Configuration;
 
-final class VueViewFinder
+class LazyComponentsResolver implements ComponentsResolver
 {
     protected const DEFAULT_DEPTH = 20;
```
Tofandel, line 9: The only thing I would change here is making the depth configurable or maybe add type `int` to `$deep` in `loadModules` and `loadModule` and use that as the `depth` param

packages/laravel/src/Http/Middleware.php
```diff
@@ -108,7 +108,7 @@ public function version(Request $request): \Closure|string|false|null
             return resolve($vite)->getHash();
```
Tofandel, line 108: Seems this is also missing a slash (the line before but I couldn't comment on it)

PR #129, opened 2024-03-20
fix(use-property): handle null values

Tofandel:
When using optional data with spatie then that value is given as null in hybridly (and not undefined)

```
props: {
  user: null
}
```

When using `useProperty('user.id')` in this case, getByPath throws an exception that it can't convert `null` to `object`

Instead of erroring we should just return undefined

This error does not happen if user was undefined
```
props: {
  user: undefined
}
```

### DIFF
--- a/packages/vue/package.json
+++ b/packages/vue/package.json
@@ -47,7 +47,7 @@
     "vue": "^3.2.45"
   },
   "dependencies": {
-    "@clickbar/dot-diver": "^1.0.4",
+    "@clickbar/dot-diver": "^1.0.5",
     "@hybridly/core": "workspace:*",
     "@hybridly/utils": "workspace:*",
     "@vue/devtools-api": "^6.6.1",
--- a/packages/vue/src/composables/property.ts
+++ b/packages/vue/src/composables/property.ts
@@ -20,14 +20,8 @@ export function useProperty<
 	path: [Override] extends [never]
 		? P
 		: string,
-): ComputedRef<ReturnType | undefined> {
-	return computed(() => {
-		try {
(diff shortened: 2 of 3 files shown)


innocenzi:
Hey @saibotk, do you think this should be fixed upstream?

saibotk:
Oh jeah definitely. 

@djfhe and I can take a look at this in https://github.com/clickbar/dot-diver directly

Tofandel:
https://github.com/clickbar/dot-diver/blob/main/src/index.ts#L245C14-L245C36

This is because `typeof null === 'object'` I've been bitten so many times by this in javascript

saibotk:
Ohhh i see, thanks for directly contributing to upstream, really appreciate it!
We'll get this in asap.

saibotk:
Info: We've released dot-diver 1.0.5, which fixes this. Huge thanks!

innocenzi:
Thank you both! Updated the PR to upgrade dot-diver 👍

saibotk:
Nice, oh what happened with the Global props type, seems like we changed something there that needs adjustment :face_with_spiral_eyes:

Tofandel:
From what I see it's the change I made from `never` to `string | number | symbol`

For some very strange reason 

`
GlobalHybridlyProperties {
	[key: string | number]: Property
}
`

Is not assignable to

`Record<string | number | symbol, unknown>`

Tofandel:
Ok, we just need to change it to `type SafeObject = Record<any, unknown>` and it fixes it without causing type issues in the lib, I really don't know why ts can't match the types because they definitely are compatible

PR #130, opened 2024-03-20
ci: update actions & nodejs version

saibotk:
This updates the GitHub action versions to their latest version. None of these had any breaking changes and were mostly adjusted to run on the latest node version.

We also fix the provenance with upgrading to node v20 in the `release` workflow. Since node v16 did contain a too old npm version it just ignored the unknown provenance setting and did not work. This is fixed now.

Additionally the explicit pnpm version in the `release` workflow was dismissed, since it will automatically choose the version as specified in the root package.json. This way we do not need to define this at multiple places and keep it in sync.

For the `tests.yml` workflow, we also added node v20 to the test matrix to cover the newest LTS. Note that v18 is kept, because it still receives security patches and is useable.

### DIFF
--- a/.github/workflows/monorepo-split.yml
+++ b/.github/workflows/monorepo-split.yml
@@ -15,7 +15,7 @@ jobs:
           - laravel
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - id: previous-tag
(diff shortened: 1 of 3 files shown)


innocenzi:
Thank you! I removed Node 18, I don't think it's necessary to keep it.

PR #133, opened 2024-03-22
feat(authorization): support tables and custom creation methods

owenconti:
Hope this helps to resolve the issue of `authorization` missing when using the `fromModel` method. Right now we're having to use `fromModel` for all of our `Data` objects due to this issue: https://github.com/spatie/laravel-data/pull/410#issuecomment-1944121984 (although maybe we should pursue a custom `ModelNormalizer` for our project 🤔)

### DIFF
--- a/docs/guide/authorization.md
+++ b/docs/guide/authorization.md
@@ -127,3 +127,43 @@ public function show(Chirp $chirp)
     ]);
 }
 ```
+
+## Using custom creation methods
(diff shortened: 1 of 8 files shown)


innocenzi:
Thanks! I added opt-out support for authorizations on tables, and opt-in support for custom creation methods. Only the latter is documented right now.

PR #134, opened 2024-03-22
fix(tables): ensure filters are applied when executing bulk actions

owenconti:
- Updates the `executeBulkAction` on the frontend to send along the currently applied filters inside the POST body
- Update the `InvokeActionController` to pull and use the refined query from the table instead of the table's model's generic query. This will automatically apply the supplied filters from the POST body.

I couldn't find any tests for these pieces, but please point me in the right direction if there are some.


innocenzi:
Looks pretty good. Will review next week and add test myself. Thanks!

innocenzi:
Thanks Owen!

PR #136, opened 2024-03-25
fix: Fail global types command and output error if TS transformer fails

owenconti:
Currently, if the TypeScript transformer command fails, the error is silently discarded and no `.php-types.d.ts` file is generated:

Screenshot 2024-03-25 at 14 22 23

This PR will monitor the output of the command and fail the `GenerateGlobalTypesCommand` if the TS transformer command fails. It will also report the output from the TS transformer command:

Screenshot 2024-03-25 at 14 20 28

### DIFF
--- a/packages/laravel/src/Commands/GenerateGlobalTypesCommand.php
+++ b/packages/laravel/src/Commands/GenerateGlobalTypesCommand.php
@@ -89,9 +89,17 @@ class GenerateGlobalTypesCommand extends Command
     protected function writeTypes(): bool
     {
         if (class_exists(TypeScriptTransformCommand::class)) {
-            Artisan::call(TypeScriptTransformCommand::class, [
+            $exit = Artisan::call(TypeScriptTransformCommand::class, [
                 '--output' => '../.hybridly/php-types.d.ts',
             ]);
+            if ($exit !== 0) {
+                $this->components->error(sprintf(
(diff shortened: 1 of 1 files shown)


nikuscs:
Thanks, this also happened to me shouldnt the definition after that be included as well?  Because Global types can also fail

owenconti:
@nikuscs the global types call is wrapped in `rescue` currently. I wasn't sure if that was intentional to cover a case where the types are expected to fail. Maybe @innocenzi can weigh in? I think all we would need to do is remove that `rescue()` wrapper.

innocenzi:
Yes, this method can fail. It's expected to recover from an exception so that we can still write `global-types.d.ts` with the default interface.

@owenconti is the end goal of this PR to make it easier to debug?

owenconti:
@innocenzi yes, exactly. For us, the `php-types.d.ts` is much more valuable than the `global-types.d.ts` since the global types on include the shared data but the PHP types include all of our custom data objects.

So we had a team member that couldn't import one of our custom data classes. Turns out `php-types.d.ts` wasn't generating even though the CLI output looked fine.

owenconti:
@innocenzi another option here would be to output the error like I have, but don't return early from the command. That way `global-types` and the rest of the command still run, but the user can see there was an issue with the PHP types generation.

PR #138, opened 2024-03-26
fix(tables): take `withoutResolvingAuthorizations` into account on data objects using custom `fromModel`

owenconti:
First off, I apologize if I am using the new `withoutAuthorizations()` incorrectly.

When using  the `withoutAuthorizations()` on a table that has a `Data` record that **does not** use a custom `fromModel` method, the authorizations are not being excluded:

```php
$table = IndexAlertTable::make()->withoutResolvingAuthorizations();
``` 

I believe this is happening because the abstract `DataResource` includes the `authorization` key by default via the pipeline:

```php
    public static function pipeline(): DataPipeline
    {
        return parent::pipeline()->firstThrough(ResolveAuthorizationsPipe::class);
    }
```

The application of the `authorization` key within the `$this->resolvesAuthorizations()` check is redundant because:

1. If the `Data` class does not have a `fromModel` then the default pipeline will add the `authorization` key.
2. If the `Data` class does have a custom `fromModel` then the user must manually add the `authorization` key via `resolveAuthorizationArray`.

I _think_ maybe what you would want to do is replace the `$this->resolvesAuthorizations()` check with the inverse and then exclude the authorizations if they are to not be resolved.

Our current workaround for this is to override the `resolveDataRecord` method in our tables and exclude the `authorization` key:

```php
    protected function resolveDataRecord(Model $model): Data
    {
        return $this->data::from($model)->excludePermanently('authorization');
    }
```

### DIFF
--- a/packages/laravel/src/Tables/Concerns/RefinesAndPaginateRecords.php
+++ b/packages/laravel/src/Tables/Concerns/RefinesAndPaginateRecords.php
@@ -162,6 +162,8 @@ trait RefinesAndPaginateRecords
                 $record->additional([
                     'authorization' => fn () => resolve(AuthorizationArrayResolver::class)->resolve($model, $this->data),
                 ]);
+            } else {
+                $record->excludePermanently('authorization');
             }
 
             return $record->toArray();
--- a/packages/laravel/src/Tables/Concerns/RefinesAndPaginateRecords.php
+++ b/packages/laravel/src/Tables/Concerns/RefinesAndPaginateRecords.php
@@ -5,7 +5,6 @@ namespace Hybridly\Tables\Concerns;
 use Hybridly\Refining\Contracts\Refiner;
 use Hybridly\Refining\Refine;
 use Hybridly\Support\Configuration\Configuration;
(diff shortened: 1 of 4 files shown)


innocenzi:
Your second point is not exactly true: using tables, it's possible to add `authorization` automatically (as you can see in the current implementation).

The question is: do we want that? Is that confusing? Is it preferable to only allow the use of `resolveAuthorizationArray`?

Right now, when using that, the `additional` call will replace the manually-added `authorization` key, which might be inefficient.

I updated the PR with a fix that doesn't break that feature, but I'm okay with removing it if you think we should.

owenconti:
@innocenzi Ah I wasn't sure if that was intentional. IMO, if I use a custom `fromModel` and want to use that `Data` instance anywhere outside of a table, then I'd have to add the `authorization` myself in `fromModel`, right?

If that's the case then I don't see the need for the Hybridly table code to try to add it for me.

innocenzi:
Okay, I agree. Let's do it this way. This is a slight breaking change though, hopefully the feature isn't used by anyone yet. Otherwise, sorry, this is going to be in a minor release. 😅

PR #139, opened 2024-03-31
fix(vue): allow overriding all form options

Tofandel:
Allows submitWith to override reset, updateInitials, timeout and transform as well and not just HybridRequestData


innocenzi:
Thanks!

PR #141, opened 2024-04-11
fix(vue): improve reactivity consistency

Tofandel:
The issue being that state.context.value = unref(context) is always a reference to the same object, so updates are never actually happening, we can fix this using

`triggerRef(state.context)`

I also optimised so that the computed properties in `useProperty` only update on viewSwap and not on each context change

### DIFF
--- a/packages/core/test/utils.ts
+++ b/packages/core/test/utils.ts
@@ -5,6 +5,7 @@ import { HttpResponse } from 'msw'
 import type { RouterContext, RouterContextOptions } from '../src/context'
 import { HYBRIDLY_HEADER } from '../src/constants'
 import { initializeContext } from '../src/context'
+import { createRouter } from '../src/router/router'
 import type { HybridPayload } from '../src/router'
 import { http } from './server'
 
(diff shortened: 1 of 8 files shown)


Tofandel:
According to https://github.com/vuejs/core/issues/10046

We can't rely on triggerRef for one particular case and that is if we use `state.context.value` in a computed (on the repro with 3.4.21 it updates reactivity exactly once and then stops)

It still works fine for nested access in computed though because then we are accessing a sub ref which has been correctly triggered as changed


You can comment `triggerRef` on the following to see that it's still doing it's job

https://play.vuejs.org/#eNp9VG1v0zAQ/iunaNJSFKUb41PVDjG0D0MaQwM+EcS8xEndOXbkl24oyn/nbOetBVZVqnuvz909d230oWnSvaXRKlrrXLHGgKbGNsCJqDZZZHQWXWYqEwCsbqQy0IIUV7SUin4XtbTCJCi4dQ9aJKBomYBRrKqounfvXNaNRRV0UCpZwykmO3UBQ9BcCo0pDTEUNtAGoRcb+mJWLl7cQinlCs6gWwR9N/dlaKkE4c598pt8gnXwKK3IDZPC1RgXxJDFCvaSFVPi5RIQIFykF8AMPEv1hNB3FvOUTNAEFe+AaSAgJFTy/eB297ijuUmf6G8d4qbYoGuSb+MYZQvYXE4pALAx3Jk7FAP8tIeeADok4IL8wNfPvmQsenzNmqQlpymXVfzgHFZw0rpfTC67ZOyMEx+n6U3Gfp20fgajek+4pd7oYeGa8sw4B0wEmtQUhK0fqdJTYyGM8NAdJ3Kc1oUyW2wgfoU0A1OYqIAIsA2ip0PEiUXxQfA5CeZEGFJsRs7Fse/8fyrzcUIATnseoRL9heU8KEZm96HGIc6scYo3/b8hIc61J+AtMdu05FKq2D8VEYWs0ewNnJ/hZ4FzTfxzKGsG62jR/sJQQqAPZl7M+ZVzStSIaTQZmXScab0Mq+8X3QsMrRuOXetXf12w/eW0W+OeQdetl04Xohx4RQkeD7QrWZXutBR4YVoXK4vceBin6q5xu4gHBkP2zI4I5/L5k5cZZWkyyPMtzZ/+Id/pFyfLoi+KaiyTZtGoM0RV1AT19dfPiHemrGVhOVq/orynuF7WYQxmV1YUCHtm59He+LuIFP6mr18MFXooygH1jfb2WYSH5eMrpU9w8foEhosOu/hrj6uG1thAvD7p2/Oo+wMaquwX

We don't use any computed with state.context anymore (because we now target state.properties in the PR) so it's worth considering wether it's better to clone the context on navigation instead so that external context use doesn't have reactivity issue either https://discord.com/channels/1032376983214116914/1133879191188799550/threads/1194130428605509643

PR #142, opened 2024-04-12
feat(form): add method to reset submission state

Tofandel:
There is currently no way to reset `form.(recently)successful` or `form.(recently)failed`

I believe they should reset automatically when the form is reset

As such I changed the default behavior of `reset` to reset those if no keys are passed

And to also clear the errors of all the fields that are reset

### DIFF
--- a/packages/vue/src/composables/form.ts
+++ b/packages/vue/src/composables/form.ts
@@ -102,18 +102,18 @@ export function useForm<
 	}
 
 	/**
-	 * Resets the form to it's initial state (including errors and flags).
+	 * Resets the fields, errors and submission state.
 	 */
-	function fullyReset() {
+	function reset() {
 		resetSubmissionState()
 		clearErrors()
-		reset()
+		resetFields()
 	}
 
 	/**
-	 * Resets the form to its initial values.
+	 * Resets the fields to their initial values.
 	 */
-	function reset(...keys: P[]) {
+	function resetFields(...keys: P[]) {
 		if (keys.length === 0) {
 			keys = Object.keys(fields) as P[]
 		}
@@ -188,7 +188,7 @@ export function useForm<
 						setInitial(fields)
 					}
 					if (optionsWithOverrides.reset !== false) {
-						reset()
+						resetFields()
 					}
 					successful.value = true
 					recentlySuccessful.value = true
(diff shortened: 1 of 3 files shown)


packages/vue/src/composables/form.ts
```diff
@@ -88,14 +88,24 @@ export function useForm<
 		})
 	}
 
+	function resetSubmissionState() {
```
innocenzi, line 94: This should have a comment to clarify the purpose, despite the clear name.

packages/vue/src/composables/form.ts
```diff
@@ -88,14 +88,24 @@ export function useForm<
...
+		successful.value = failed.value = recentlyFailed.value = recentlySuccessful.value = false
```
innocenzi: These should be one assignation per line, please.

packages/vue/src/composables/form.ts
```diff
@@ -88,14 +88,24 @@ export function useForm<
...
+		clearTimeout(timeoutIds.recentlySuccessful!)
+		clearTimeout(timeoutIds.recentlyFailed!)
+		progress.value = undefined
+	}
+
 	/**
 	 * Resets the form to its initial values.
 	 */
 	function reset(...keys: P[]) {
```
innocenzi: I disagree with the changes to `reset`. This function should reset `fields` and that's it. I'm open to discuss having another function to reset everything though.

packages/vue/src/composables/form.ts
```diff
@@ -165,7 +171,6 @@ export function useForm<
 					return hooks.error?.(incoming, context)
 				},
 				success: (payload, context) => {
-					clearErrors()
```
innocenzi: Is this on purpose? Errors should be cleared on success.

innocenzi (changes requested):
I'm skeptical about this one, mainly due to the `reset` changes. 

I also don't see how useful it can be to manually reset `recentlySuccessful` and `recentlyFailed`, though I'm open to having a way to do so.

packages/vue/src/composables/form.ts
Tofandel: They were by reset, but this is now in a different method

Tofandel:
It's not so much about `recentlySuccessful` but more about `failed` and `successful` I'm using them to hide the form and show a validation message after a form has been submitted

But I want to provide a button to manually reset and allow the form to be resubmitted (remember the form is hidden via `form.successful`), a call to reset currently doesn't reset `successful` or `failed` and thus the form would always be hidden hence the motivation for this PR

PR #147, opened 2024-04-25
feat(vite): support ddev and lando environments

MichelJonkman:
With this PR the Vite plugin will detect a `.ddev` folder or a `.lando.yml` file and set a prefix to the php executable unless `PHP_EXECUTABLE_PATH` is set. 

Includes some extra messages when an error occurs during `php artisan hybridly:config` because running Vite without starting ddev/lando first will add a lot of extra messages to the output of the command and cause an error to happen on startup.

Also includes `Detected dev environment: ddev` in the console output below `Registered` and above `Version mismatch` to indicate what environment was found. This is to clarify to the developer that it's not using the default php path and hopefully prevent confusion.

Technically this could break some environments for people if they don't use ddev/lando but do have the folder/file in the project, but this is very unusual as these should not exist in the project if a developer is not using them.

### DIFF
--- a/packages/vite/src/config/env.ts
+++ b/packages/vite/src/config/env.ts
@@ -1,23 +1,26 @@
 import fs from 'node:fs'
 import { loadEnv } from 'vite'
 
+let phpExecutable: string[]
+let devEnvironment: 'ddev' | 'lando' | 'native'
+
+/**
+ * Gets all environment variables, including `.env` ones.
+ */
 export function getEnv() {
 	return { ...process.env, ...loadEnv('mock', process.cwd(), '') }
 }
 
-let phpExecutable = null
-
-export function getPhpExecutable(): string[] {
+export async function getPhpExecutable(): Promise<string[]> {
 	if (phpExecutable) {
 		return phpExecutable
 	}
 
 	const env = getEnv()
-
 	const php = (env.PHP_EXECUTABLE_PATH ?? 'php').split(' ')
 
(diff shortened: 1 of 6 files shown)


innocenzi:
@MichelJonkman just cleaned up the code a bit, can you double check everything works for you? 

Only thing I'm worried about is the `fs.existsSync` calls that might slow down HMR initially, until the result is cached. 

> Technically this could break some environments for people if they don't use ddev/lando but do have the folder/file in the project, but this is very unusual as these should not exist in the project if a developer is not using them.

This could happen in a team where people have native environments and others prefer using Docker. If we get reports about that, we can follow-up with a solution.

MichelJonkman:
@innocenzi 

> Only thing I'm worried about is the fs.existsSync calls that might slow down HMR initially, until the result is cached.

That's the reason why I chose to cache it, but thinking about it it shouldn't be much of an issue compared to the amount of filesystem calls that happen just from starting the dev server and then running 3 PHP commands that each have quite a lot of filesystem calls to do. Unless node is exceptionally slow with checking if a file exists.

> This could happen in a team where people have native environments and others prefer using Docker. If we get reports about that, we can follow-up with a solution.

It would be really easy to add an environment variable to force the detected environment in that case. 

I'll have a look if things still work in a bit!

MichelJonkman:
@innocenzi Seems to work just fine, sorry about the incorrect Typescript stuff, PhpStorm doesn't wanna run my language servers anymore 🙃

innocenzi:
Thank you!

PR #150, opened 2024-04-29
fix(laravel): flush shared properties on octane requests

### LINKED ISSUE #148: [Laravel Octane] Shared data between middleware seem to leak with one another
### Describe the bug

When having multiple HandleHybridRequests, it seems that the variables between the two middleware leak into each other.

An example when to need multiple Hybridly middleware is when having multiple route groups. ie: Marketing/Admin/Control Panel/App/Staff which will have different global properties.

### Reproduction

https://github.com/jaggy/hybridly-octane-issue

### Steps to reproduce

- Start Octane with Swoole
  - Seems to work with Roadrunner.
- Visit /one
  - This will show the correct global properties filled. Only `one` and `lazyOne` will be filled.
- Visit /two
  - This will have `two` and `lazyTwo filled, but ALSO `one` and `lazyOne` which are only in the previous middleware group.

Here's a video of the issue

https://github.com/hybridly/hybridly/assets/1993075/3e6302dd-9269-425a-bd51-311672a42865

### System information

```bash
System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 256.11 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Yarn: 3.3.1 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - ~/Library/Application Support/Herd/config/nvm/versions/node/v20.11.1/bin/npm
    pnpm: 7.19.0 - ~/Library/pnpm/pnpm
    bun: 1.0.33 - ~/.bun/bin/bun
  Browsers:
    Chrome: 124.0.6367.91
    Safari: 17.4.1
    Safari Technology Preview: 17.4
  npmPackages:
    hybridly: ^0.7.7 => 0.7.7
```

### Used package manager

npm

### Logs

_No response_

### Validations

innocenzi:
Closes #148

### DIFF
--- a/packages/laravel/src/HybridlyServiceProvider.php
+++ b/packages/laravel/src/HybridlyServiceProvider.php
@@ -26,6 +26,7 @@ use Illuminate\Support\Facades\Route;
 use Illuminate\Testing\TestResponse;
 use Illuminate\View\Compilers\BladeCompiler;
 use Illuminate\View\Factory;
+use Laravel\Octane\Events\RequestReceived;
 use Spatie\LaravelPackageTools\Package;
 use Spatie\LaravelPackageTools\PackageServiceProvider;
 
@@ -56,6 +57,7 @@ class HybridlyServiceProvider extends PackageServiceProvider
     public function bootingPackage(): void
     {
         $this->registerActionsEndpoint();
+        $this->registerOctaneListener();
     }
 
     public function packageBooted(): void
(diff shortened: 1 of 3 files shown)


packages/laravel/src/HybridlyServiceProvider.php
```diff
@@ -80,6 +82,18 @@ public function packageBooted(): void
...
     }
 
+    protected function registerOctaneListener(): void
+    {
+        if (!class_exists(RequestReceived::class)) {
+            return;
+        }
+
+        $this->app['events']->listen(
+            RequestReceived::class,
```
RobertBoes: Think you'd also need to do this for `TaskReceived::class` and `TickReceived::class`

innocenzi: Thanks!

PR #151, opened 2024-05-06
feat: useResource

MichelJonkman:
I created a rough proof of concept for my `useResource` idea, this name is not final of course I just think it fits well. 

Under the hood it's basically an API endpoint that fetches simple data, this means it doesn't accept parameters by design. It then caches what route name the the resource was requested on and uses that data to preemptively get that resource and pass it to the Hybridly frontend.

Some changes that need to be made are probably the way they're cached, maybe it's possible to build a smarter solution, and there should be a way to fetch resources without it remembering in case a resource is expensive to request or the resource is not always used on the page.

Typescript is also not implemented yet as I'm not experienced enough with both Typescript definitions and `typescript-transformer`. My idea for this however, was to use docblocks as seen below.

```php
/** @return UserData[] */
hybridly()->registerResource('users', function() {
    return UserData::collect(User::all());
});
```

`useResource` returns a ref with the given default and gets updated as soon as the resource has resolved, this is instant when resource was cached on the route.

```ts
const users = useResource('users', []);
```

There are likely many things wrong with my integration into Hybridly and Typescript will probably whine at a lot of things but I thought I'd first show off the idea before spending too much time on it!

### DIFF
--- a/packages/laravel/src/Hybridly.php
+++ b/packages/laravel/src/Hybridly.php
@@ -16,6 +16,7 @@ final class Hybridly
     use Concerns\HasSharedProperties;
     use Concerns\HasVersion;
     use Concerns\ResolvesUrls;
+    use Resources\Concerns\HasResources;
     use Conditionable;
     use ForwardsCalls, Macroable {
         Macroable::__call as macroCall;
--- a/packages/laravel/src/HybridlyServiceProvider.php
+++ b/packages/laravel/src/HybridlyServiceProvider.php
@@ -10,6 +10,7 @@ use Hybridly\Commands\InstallCommand;
 use Hybridly\Commands\MakeTableCommand;
 use Hybridly\Commands\PrintConfigurationCommand;
 use Hybridly\Http\Controller;
+use Hybridly\Resources\Http\ServeResourceController;
 use Hybridly\Support\Configuration\Configuration;
 use Hybridly\Support\Data\PartialLazy;
 use Hybridly\Support\RayDumper;
@@ -204,6 +205,10 @@ class HybridlyServiceProvider extends PackageServiceProvider
             Route::post($this->getConfiguration()->tables->actionsEndpoint, InvokedActionController::class)
                 ->middleware($this->getConfiguration()->tables->actionsEndpointMiddleware)
                 ->name($this->getConfiguration()->tables->actionsEndpointName);
+
+            Route::post('/hybridly/resource', ServeResourceController::class)
+                ->middleware('web')
+                ->name('hybridly.resource');
         }
     }
 
--- a/packages/laravel/src/Resources/Concerns/HasResources.php
+++ b/packages/laravel/src/Resources/Concerns/HasResources.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Hybridly\Resources\Concerns;
+
+use Exception;
+
+trait HasResources
+{
+    /** @var array<string, callable> */
+    protected array $registeredResources = [];
+
+    public function registerResource(string $key, callable $callback): void
+    {
+        $this->registeredResources[$key] = $callback;
+    }
+
+    /**
+     * @throws Exception
+     */
+    public function getResource(string $key): mixed
+    {
+        if (!$this->hasResource($key)) {
+            throw new Exception('Resource does not exist');
+        }
+
+        return $this->registeredResources[$key]();
+    }
+
+    public function hasResource(string $key): bool
+    {
+        return isset($this->registeredResources[$key]);
+    }
+}
--- a/packages/laravel/src/Resources/Http/ServeResourceController.php
+++ b/packages/laravel/src/Resources/Http/ServeResourceController.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Hybridly\Resources\Http;
+
+use Exception;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Cache;
+
+class ServeResourceController
+{
+    /**
+     * @throws Exception
+     */
+    public function __invoke(Request $request)
+    {
+        $payload = $request->json();
+        $resources = [];
+
+        $keys = $payload->all('keys');
+
+        foreach ($keys as $key) {
+            $resources[$key] = hybridly()->getResource($key);
+        }
+
+        if ($route = $payload->get('route')) {
+            Cache::set('hybridly.resources.routes.' . $route, $keys, 3600);
+        }
+
+        return json_encode($resources);
+    }
+}
--- a/packages/laravel/src/Resources/Http/ServeResourcesMiddleware.php
+++ b/packages/laravel/src/Resources/Http/ServeResourcesMiddleware.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Hybridly\Resources\Http;
+
+use Closure;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Cache;
+
+class ServeResourcesMiddleware
+{
+    public function handle(Request $request, Closure $next)
+    {
+        $routeName = $request->route()->getName();
+
+        $resourceKeys = Cache::get('hybridly.resources.routes.' . $routeName);
+        if($resourceKeys) {
+            $hybridly = hybridly();
+            $resources = [];
+
+            foreach ($resourceKeys as $resourceKey) {
+                if(!$hybridly->hasResource($resourceKey)) {
+                    continue;
+                }
+
+                $resources[$resourceKey] = $hybridly->getResource($resourceKey);
+            }
+
+            $hybridly->share('resources', $resources);
+        }
+
+        return $next($request);
+    }
+}
--- a/packages/vue/src/composables/resource.ts
+++ b/packages/vue/src/composables/resource.ts
@@ -1,0 +1,59 @@
+import axios from "axios";
+import {Ref} from "vue";
+import {state} from "../stores/state";
+
+let queued = {};
+let refs = {};
+
+let queueTimeout = null;
+
+export function useResource<T>(key: string, placeholder: T): Ref<T> {
+	if (refs.hasOwnProperty(key)) {
+		return refs[key];
+	}
+
+	const resource = ref<T>(placeholder);
+
+	queued[key] = refs[key] = resource;
+	initiateResolve();
+
+	return resource;
+}
+
+function initiateResolve() {
+	if (queueTimeout) {
+		return;
+	}
+
+	queueTimeout = setTimeout(() => {
+		queueTimeout = null;
+		resolveResources();
+	}, 50);
+}
+
+async function resolveResources() {
+	const keys = Object.keys(queued);
+	queued = {};
+
+	const response = await axios.post(route('hybridly.resource'), {
+		keys: keys,
+		route: router.current()
+	});
+
+	for (const [key, value] of Object.entries(response.data)) {
+		refs[key].value = value;
+	}
+}
+
+export function initResources() {
+	queued = {};
+	refs = {};
+	queueTimeout = null;
+
+	if (state.properties?.resources) {
+		for (const [key, value] of Object.entries(state.properties?.resources)) {
+			refs[key] = ref(value);
+		}
+	}
+}
+
--- a/packages/vue/src/initialize.ts
+++ b/packages/vue/src/initialize.ts
@@ -11,6 +11,7 @@ import { devtools } from './devtools'
 import { dialogStore } from './stores/dialog'
 import { onMountedCallbacks } from './stores/mount'
 import { viewTransition } from './plugins/view-transition'
+import {initResources} from "./composables/resource";
 
 /**
  * Initializes Hybridly's router and context.
@@ -39,6 +40,7 @@ export async function initializeHybridly(options: InitializeOptions = {}) {
 			},
 			onContextUpdate: (context) => {
 				state.setContext(context)
+				initResources()
 			},
 			onViewSwap: async(options) => {
 				if (options.component) {


innocenzi:
Hey! I see the idea, but I'm wondering how different this would be from simply using global properties (lazily).

Could you perhaps show me use cases so I can get a clearer idea?

MichelJonkman:
@innocenzi I would use it mostly for populating select lists, it is similar to lazy loaded global properties but with these I'm trying to reduce the amount of requests to the server while keeping a good dev experience.

Imagine you have a form with 6 different lists, all lazy loaded and used by people a lot, it creates a lot unnecessary network requests. This could be solved by passing the data in the controller but it's easy to forget something and harder to use when working in teams. Using this feature it all be handled in the background and would reduce bugs.

This could also be added as a feature to lazy global properties but this was easier to create as a proof of concept.

innocenzi:
How about solving this with headless frames? 👀 

```vue
<hybrid-frame
	v-slot="{ users }: { users: App.Data.UserProfileData[] }"
	:href="route('users.list')"
>
  <input-select :users="users" />
</hybrid-frame>
```

innocenzi:
I built a stupid proof-of-concept using frames. The only issue is caching right now, but this could be added, as well as a programmatic way of loading frames. Essentially, we'd have a way of making an Hybridly-flavored AJAX request. What do you think?


https://github.com/hybridly/hybridly/assets/16060559/8aa5ccff-ee90-446f-96f8-1e76e1670043

<details>
<summary>See code</summary>

```vue
<script setup lang="ts">
import {
  Listbox,
  ListboxLabel,
  ListboxButton,
  ListboxOptions,
  ListboxOption,
} from '@headlessui/vue'

const selected = ref<App.Data.ChirpData>()
</script>

<template>
  <div class="">
    <Listbox v-model="selected">
      <div class="relative mt-1">
        <ListboxButton
          class="relative w-full cursor-default rounded-lg bg-white py-2 pl-3 pr-10 text-left shadow-md focus:outline-none focus-visible:border-indigo-500 focus-visible:ring-2 focus-visible:ring-white/75 focus-visible:ring-offset-2 focus-visible:ring-offset-orange-300 sm:text-sm"
        >
          <span class="block truncate">{{ selected?.body ?? 'No chirp selected' }}</span>
          <span
            class="pointer-events-none absolute inset-y-0 right-0 flex items-center pr-2"
          >
            <ChevronUpDownIcon
              class="h-5 w-5 text-gray-400"
              aria-hidden="true"
            />
          </span>
        </ListboxButton>

        <transition
          leave-active-class="transition duration-100 ease-in"
          leave-from-class="opacity-100"
          leave-to-class="opacity-0"
        >
          <ListboxOptions
            class="absolute z-40 mt-1 max-h-60 w-full overflow-auto rounded-md bg-white py-1 text-base shadow-lg ring-1 ring-black/5 focus:outline-none sm:text-sm"
          >
						<hybrid-frame :href="route('chirps.list')" v-slot="{ chirps }: { chirps: App.Data.ChirpData[] }">
							<ListboxOption
								v-slot="{ active, selected }"
								v-for="chirp in chirps"
								:key="chirp.id"
								:value="chirp"
								as="template"
							>
								<li
									:class="[
										active ? 'bg-amber-100 text-amber-900' : 'text-gray-900',
										'relative cursor-default select-none py-2 pl-10 pr-4',
									]"
								>
									<span
										:class="[
											selected ? 'font-medium' : 'font-normal',
											'block truncate',
										]"
										>{{ chirp.body }}</span
									>
									<span
										v-if="selected"
										class="absolute inset-y-0 left-0 flex items-center pl-3 text-amber-600"
									>
										<CheckIcon class="h-5 w-5" aria-hidden="true" />
									</span>
								</li>
							</ListboxOption>
						</hybrid-frame>
          </ListboxOptions>
        </transition>
      </div>
    </Listbox>
  </div>
</template>

```

</details>

MichelJonkman:
Does this cache same way though? In my proof of concept I cache that a route uses the resource and then pass that resource on page load instead of lazy

MichelJonkman:
Note that I also defer requests and group them together, forgot about that feature. So if 6 different resources are requested within I think 50ms rn, they get grouped and sent with 1 request.

innocenzi:
It doesn't cache at all for now. And no request bundling either. Honestly, I like the idea, but I don't want too many of these feature. I think frames and this have the potential to be factorized into the same API.

nikuscs:
The PR and feature itself is cool indeed, but like frames, tables, etc i believe they should probably be “extracted” from the core package into their own packages/plugins, there is a ton of cool ideias that each project needs, i believe Hybridly already provides a lot of good DX and the code base should be as simple as possible, otherwise if we ever imagine on creating adaptors for other frameworks for example, the amount of work will keep just raising, raising the bar for people who might even consider contributing. 

Also, this goes a bit “out” of the page concept, i can see this leading probably to more confusion as data is shared across multiple different places.
Once again excellent work!

MichelJonkman:
@innocenzi Either frames or lazy global properties could be expanded with the same features, I haven't used frames yet so I'm not sure how they work. If these features are added to frames it would be very useful to have a `useFrame()` function for the same effect as my `useResource()`, the rest of the features can then be added as configuration.

From your example video a big difference I'm noticing is that the data only loads when you click on it, this could be fixed by adding an eager loading option though.

I agree on not having too many different features with overlap though!

MichelJonkman:
@nikuscs I agree that features like this should be part of a separate package but it should still be an official package imo, just optional if you don't want these features. For now it's a lot easier to manage if they're part of the main package, especially since it only support Vue for now.

The only confusion I can see over what is already implemented is that it has overlap and people might think "But why not use [other feature] instead?", which is still a valid issue. Combining it with an existing feature would be my preference.

Thank you!

MichelJonkman:
Another feature `useResource` would have if I knew how to it worked is automatic typing, so you wouldn't have to use `{ chirps: App.Data.ChirpData[] }` as in the example. I'm not sure if that could be added to frames as well?

innocenzi:
Valid point regarding the package separation, but a Hybridly package like this is annoying to maintain (it has to be hosted on both npm and packagist). 

A first-party package doesn't make sense to me, it's also more maintenance burden (also new packages to host and build). If we add this, it's going to be built-in: for PHP it doesn't matter, and for JS it will be tree-shaken.

The only real issue is feature duplication, and if we want this, we have to work out a better API

innocenzi:
@MichelJonkman regarding automated typing, it would mean resource declaration would have to be static (as-in, not loaded dynamically) in order to generate interfaces from them (like we do for the middleware)

MichelJonkman:
> it would mean resource declaration would have to be static (as-in, not loaded dynamically) in order to generate interfaces from them (like we do for the middleware)

What do you mean with this exactly?

innocenzi:
> What do you mean with this exactly?

Basically that the `hybridly:types` artisan command *has* to have access to all resources. If a resource is specified in a condition (eg. based on the logged in user), then that artisan command won't know about it and types won't be generated.

MichelJonkman:
> > What do you mean with this exactly?
> 
> Basically that the `hybridly:types` artisan command _has_ to have access to all resources. If a resource is specified in a condition (eg. based on the logged in user), then that artisan command won't know about it and types won't be generated.

Oh yes, I guess it wasn't very clear but the resources are meant to be registered globally in a ServiceProvider. Of course this can be changed to something else but the idea would be the same.

Tofandel:
To me it sounds exactly like global lazy

If you want the lazy to be loaded on specific routes automatically you can use

`Lazy::when(fn () => $request->is('route'), fn () => LAZY)`

But if they need to be added to specific routes then maybe its even better to just add them to the route's properties and create a wrapper to get the resource

I don't really see the use case for this either

Tofandel:
> Valid point regarding the package separation, but a Hybridly package like this is annoying to maintain (it has to be hosted on both npm and packagist).

Not necessarily you can provide the full plugin as a npm package and have composer install from a dir inside the package, or the other way around, have a composer plugin that provides the npm package that is then installed locally

It has the advantage of always having both versions match

MichelJonkman:
@Tofandel The point of `useResource` is to provide a way to request a resource, like a list of users, from a component and remember that the page requested the resource. Then next time someone loads that page it automatically passes that resource so it doesn't have to do an asynchronous request again. It also groups together requests that are done in quick succession.

This is all meant to reduce the amount of requests to the server while keeping the same developer experience of just being able to use a component without thinking about what data you need to pass in your controller. These features could of course be implemented in global properties, the proof of concept was just a lot easier for me to build like this.

Tofandel:
With this description I still don't see how this would be beneficial

Am I understanding correctly:

For example you have a profile page, in that page you have a select that loads the list of all users when clicked, and another select that loads all the roles when clicked

You'd rather have both selects load with the rest of the properties when you first visit the page rather than with a user interaction?

If so that's the opposite of an improvement because now the initial load will take more time because the page will load things the user doesn't necessarily need, a request on user interaction is never a bad thing because a user interacts at a slow pace and have a reaction time so they have a much better time waiting 400ms for a select to load on an already loaded page than 2s for a page to load

There is also the fact that you'd have requests where this is not happening until one user does open the selects because of the cache. And after that all the requests will. That is non deterministic and to me that seems wobbly and prone to a lot more hard to debug bugs (race condition etc)

In the end this would just result in props with extra steps when you could simply have added a callback prop to your route and you would get the same effect but with the advantage of being deterministic so you control what is loaded and when, not your users clicking somewhere and deciding for other users what should be loaded, `useResource` takes this control away

MichelJonkman:
After thinking about it for a bit and looking at how it can be combined with frames, I think that would be the best solution. 

The caching in this case should only be used for data that will always be needed on a page, not when there's a chance that the data will not be used. With my previous example all selects would always be used by the user so it's better to already have the data. It's also not meant to be used when getting that data takes a lot of time or it has side effects.

The determinism is an issue that I think doesn't matter if it's handled as intended, but it is quite easy to abuse it and end up with hard to debug errors. The best way I can think of to fix this is by having a Vite plugin gather the uses of a resource on a page and tell Laravel about it but I feel like that is quite a big undertaking and while running the Vite dev server it's already done with the HTTP request by the time it starts compiling the page component.
Definitely something to think about though, I think it's best if the frames feature is modified to allow for the same functionality without the remember/cache feature and think about that more afterwards.

Tofandel:
But then what you want is just a prop that you can reuse in multiple pages, no need for anything extra, just create a class `Resources` in it add a method to retrieve the User list and a method to retrieve the role list for example and then simply go `'users' => fn () => Resources::userList()`

Is your issue about the persistence of the prop? Aka you want the prop to still be kept in memory when you change page because another page might use it? If that is the case then there is another solution that I'm using myself it's simply to provide a `useCachedProperty` that simply keeps a local cache like this


```ts
import { useProperty } from '@hybridly/vue';
import { ComputedRef } from 'vue';

const computedCache: Record<string, ComputedRef<unknown>> = {};
const scope = effectScope(true);

export default (path: Parameters<typeof useProperty>[0]): ReturnType<typeof useProperty> => {
  if (! (path in computedCache)) {
    scope.run(() => {
      const cache = ref();
      const property = useProperty(path);
      watch(property, (property) => {
        if (property !== undefined) {
          cache.value = property;
        }
      }, { immediate: true });
      computedCache[path] = computed(() => {
        return property.value === null || property.value === undefined ? cache.value : property.value;
      });
    });
  }
  return computedCache[path];
};
```

Or is it that you don't know which components are going to be on the page and you want the components to request the props? (Though I would argue that even if the components are dynamic, the backend knows which components it will request the front to display so at some points you still can know all the props your page is going to need even if it requires some processing)

MichelJonkman:
The core what I want is basically frames but easier to define/register, typed and a way of getting the frame data programmatically.  I want to avoid passing the data in the controller as it's a pitfall for other developers who have to work with it and don't know enough about how it's implemented. I have a colleague that doesn't have much experience yet and I want him to use a component without thinking about what data it should have, since for some components the data is always the same. I also want to avoid using `provide` and `inject` as I feel like those are also quite easy to make mistakes with if you need to define it on a lot of pages. 

Having it be automatically typed and registered globally also makes it much easier if you need to change the data it receives, especially when changing the `Data` class it returns to another one or changing the query it runs. 

Of course these issues can all be solved without a built in solution but I feel like it would be a lot easier and prevent mistakes.

MichelJonkman:
> Or is it that you don't know which components are going to be on the page and you want the components to request the props?

That is a big part of it yes, but more because I don't want myself or anyone else to have to think about what data the component requires and how to get it there. Especially if it's nested in 1 or more other components.

Tofandel:
> The best way I can think of to fix this is by having a Vite plugin gather the uses of a resource on a page and tell Laravel about it but I feel like that is quite a big undertaking and while running the Vite dev server it's already done with the HTTP request by the time it starts compiling the page component.

But how do you determine the dependency tree of a page? What if you have dynamic components on a page? This is an impossible task without actually running a real production request (Basically SSR)

I think you should instead make a class for each of the component that needs some data, and have the class provide it

That way you can yourself build a dependency tree of the data that is needed in your page and in the components you can just use `useProperty()`

But yes I also agree this could be a nice use case for frames, but then frames should be able to be included from the initial request which is not the case, they are only included on mounted so they would trigger a lot of requests and likely duplicated ones

Something like `return hybridly('view', [])->withFrames(['frame_url'])`

But it will never be able to deduplicate data this way, so if you have multiple frames and they need a shared data, the data would be sent for each frame instead of just once

Maybe instead you can just make a third party package using the backend map method that I explained

You could have an interface like `HasPropertyDependency` which requires a method `getDependencies` which returns an array of dependency keys that the components needs

Then define your components `class Component1 implement HasPropertyDependency`

And then in the page controller go with something like `return withComponents(hybridly('view'), [Component1::class, Component2::class])`

withComponents would merge all the dependencies, retrieve them either with a router or simply a facade and add the properties to the hybridly response

Then you components simply need `useProperty`

MichelJonkman:
> I think you should instead make a class for each of the component that needs some data, and have the class provide it

Wouldn't you still have the determinism issue if you did this? Since you still need at least 1 page load before you have that dependency tree.

Eager loading frames using `->withFrames()` is a very good solution as well, together with grouping requests I think the largest performance issues I can foresee are solved. When using a lot of frames on a page without dynamic parameters you can eager load them from the controller but it doesn't stop you from using the component including that frame anywhere else.

I suggested to Enzo in private to use an implementation similar to mine for registering frames instead of making them controllers with routes. This means that something like `->withFrames(['users'])` is possible, if they need parameters it will be a little trickier but the general idea with remain the same.

> But it will never be able to deduplicate data this way, so if you have multiple frames and they need a shared data, the data would be sent for each frame instead of just once

How so? If the frame's data is retained until you navigate to another page I don't see any reason why it would have to do multiple requests. If a frame is requested that uses the same name and parameters the same data of the previous call should be used imo, with an option to request fresh data instead.

I solved this in my implementation by storing the `ref` in a global object and returning that every time the same resource is requested. Then when it resolves it simply updates that `ref` and everything that uses it instantly has that same data.

Tofandel:
> Wouldn't you still have the determinism issue if you did this? Since you still need at least 1 page load before you have that dependency tree.

No, I meant you give the dependency tree yourself in the controller like this `return withComponents(hybridly('view'), [Component1::class, Component2::class])`

> How so? If the frame's data is retained until you navigate to another page I don't see any reason why it would have to do multiple requests. If a frame is requested that uses the same name and parameters the same data of the previous call should be used imo, with an option to request fresh data instead.


Each frame has it's own scoped data and you cannot access it from outside the frame, this means that if multiple frames use the same data, you still need to include it as many times you have frames, yes keeping data between navigation would indeed be possible, I'm talking about if say you want the list of users in Frame 1, Frame 2 and Frame 3 then each frame will need to get the list of users

MichelJonkman:
> No, I meant you give the dependency tree yourself in the controller like this `return withComponents(hybridly('view'), [Component1::class, Component2::class])`

Aaah like that, I think the `withFrames()` feature would cover that as well so I think that's my current preferred solution!

> Each frame has it's own scoped data and you cannot access it from outside the frame, this means that if multiple frames use the same data, you still need to include it as many times you have frames

I see it more as passing the data as a property to the frame, so it's readonly and can be reused however much you want. Potentially if using something like my implementation the data can also be refreshed and updated everywhere very easily. Would also need configuration options to control how it behaves of course, but this could be added in the future. 

A use case where this could be useful is if you have 2 or more select elements with different purposes but the same data that need to be refreshed for whatever reason, like with a refresh button. But it's not something I'd see myself using very often.

Tofandel:
> A use case where this could be useful is if you have 2 or more select elements with different purposes but the same data that need to be refreshed for whatever reason, like with a refresh button. But it's not something I'd see myself using very often.

That's exactly what I'm explaining wouldn't work very well with frames because of their scope, but it would work with the util that you can make because it's using the page properties which are global, I would suggest the name `withPropertiesFrom`

innocenzi:
I read your ideas—I'll close both PRs and try to come up with something more flexible than just the frames, but with them as a foundation, taking into account the ideas in this PR.

MichelJonkman:
@innocenzi Feel free to message me on Discord if you have any ideas you wanna discuss!

PR #161, opened 2024-06-29
feat(tables): inline tables

innocenzi:
This pull request adds support for rendering basic tables without a class:

```php
return view('chirps.table', [
    'users' => AnonymousTable::create(
        model: Chirp::class,
        columns: [
            Columns\TextColumn::make('id')->label('#'),
            Columns\TextColumn::make('body'),
        ],
        refiners: [
            Filters\Filter::make('body', alias: 'search')->loose(),
        ]
    )
]);
```

These will not have support for a few features, like extending query (for simplicity) or actions (because they need a parent class).


Tofandel:
Maybe `OrphanedTable` or `InlineTable` is a better name for this feature. Anonymous kinda feels like it's something to do with authentication

innocenzi:
I chose `Anonymous` because it doesn't have a name/identifier like other tables, kind of like `AnonymousEvent` or `AnonymousNotifiable` in Laravel. Tables don't have a link to authentication in any way, so I don't think there could be any confusion.

I like `InlineTable` too, not sure if I want to change the name.

iRaziul:
> I chose `Anonymous` because it doesn't have a name/identifier like other tables, kind of like `AnonymousEvent` or `AnonymousNotifiable` in Laravel. Tables don't have a link to authentication in any way, so I don't think there could be any confusion.
> 
> I like `InlineTable` too, not sure if I want to change the name.

`InlineTable` is best suited for this use case.

PR #165, opened 2024-07-08
fix(laravel): gracefully return responses from actions

### LINKED ISSUE #164: Action response can not showing flash message
### Describe the bug

Here is the code:

```
    protected function defineActions(): array
    {
        return [
            BulkAction::make('delete')
                ->action(function (Collection $records) {
                    $records->each->delete();

                    return back()->with('success', 'Success!');
                }),
            InlineAction::make('delete')
                ->action(fn (User $user) => $user->delete()),
        ];
    }
```

### Reproduction

https://hybridly.dev/guide/tables.html#action-responses

### Steps to reproduce

I tried to troubleshoot the error

It seems that the response returned by the action does not have a set-cookie Header

However, a set-cookie Header exists for responses returned by custom routes

image

image

### System information

```bash
System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 137.78 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.2.0 - /opt/homebrew/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.7.0 - /opt/homebrew/bin/npm
    pnpm: 9.3.0 - /opt/homebrew/bin/pnpm
    bun: 1.1.13 - /opt/homebrew/bin/bun
  Browsers:
    Brave Browser: 126.1.67.123
    Chrome: 126.0.6478.127
    Edge: 126.0.2592.87
    Safari: 17.5
  npmPackages:
    hybridly: ^0.7.9 => 0.7.9
```

### Used package manager

pnpm

### Logs

_No response_

### Validations

robin-dongbin:
close #164

### DIFF
--- a/packages/laravel/config/hybridly.php
+++ b/packages/laravel/config/hybridly.php
@@ -75,7 +75,7 @@ return [
         'enable_actions' => true,
         'actions_endpoint' => 'invoke',
         'actions_endpoint_name' => 'hybridly.action.invoke',
-        'actions_endpoint_middleware' => [],
+        'actions_endpoint_middleware' => ['web'],
     ],
 
     /*
--- a/packages/laravel/src/Tables/Actions/Http/InvokedActionController.php
+++ b/packages/laravel/src/Tables/Actions/Http/InvokedActionController.php
@@ -84,8 +84,7 @@ final class InvokedActionController
         );
 
         if ($result instanceof Response) {
-            $result->send();
-            exit;
+            return $result;
         }
 
         return back();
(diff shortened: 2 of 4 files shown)


innocenzi:
I think I used `send` for a reason, but I can't remember why right now. I'll try to dig in and write related tests before merging.

robin-dongbin:
Also I think `exit` is not good for laravel octane use case

innocenzi:
Definitely. Could you add tests for these? Otherwise I'll do it, but it'll take a moment until I have some free time.

innocenzi:
Found time to add tests myself actually. Thank you!

PR #167, opened 2024-07-09
feat(vue): add `selectPage` and `deselectPage`

robin-dongbin:
Just think of the feature as a common requirement, feel free to rename or change anything of this PR.

### DIFF
--- a/docs/api/utils/use-table.md
+++ b/docs/api/utils/use-table.md
@@ -16,9 +16,9 @@ function useTable<
   Properties extends Record<string, any>,
 	Paginator extends 'simple' | 'cursor' | 'length-aware',
 >(
-  properties: Properties,
-  tableKey: keyof Properties,
-  defaultOptions: HybridRequestOptions = {}
+	properties: Properties,
+	tableKey: keyof Properties,
+	defaultOptions: HybridRequestOptions = {}
 )
 ```
 
@@ -28,7 +28,7 @@ function useTable<
 
 ```ts
 const $props = defineProps<{
-  users: Table<App.Data.UserData>
+	users: Table<App.Data.UserData>
 }>()
 
 const users = useTable($props, 'users')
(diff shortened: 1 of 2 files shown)


docs/api/utils/use-table.md
```diff
@@ -255,6 +255,18 @@ Selects all records.
 
 De-selects all records.
 
+### `selectPage`
+
+- Type: `Function`
+
+Selects paged records.
```
innocenzi: Please include usage example, because this is very unclear what it does.

packages/vue/src/composables/table.ts
```diff
@@ -161,6 +161,10 @@ export function useTable<
 		selectAll: bulk.selectAll,
 		/** Deselects all records. */
 		deselectAll: bulk.deselectAll,
+		/** Selects paged records. */
+		selectPage: bulk.select(...table.value.records.map((record: RecordType) => getRecordKey(record))),
+		/** Deselects paged records. */
+		deselectPage: bulk.deselect(...table.value.records.map((record: RecordType) => getRecordKey(record))),
```
innocenzi: This seems wrong. `selectPage` and `deselectPage` are not functions, since the underlying `select` and `deselect` are called at the moment they are defined. 

```ts
const { selectPage } = useTable()

// `selectPage` is not a function. It's just `void`.
```

innocenzi (changes requested):
Can you show me the use case? Not against implementing it, but I'd like to understand what it does.

docs/api/utils/use-table.md
robin-dongbin: Most of the time, datasets are presented in pagination,

The purpose of this feature is to provide a function to select all records for the current page

innocenzi: Oh okay, the naming was unclear. This is better, thanks!

innocenzi:
Thank you!

PR #169, opened 2024-07-10
feat(dialog): allow `show` to be used as v-model

Tofandel:
By using a computed setter we can make the `show` of the dialog usable as a model value

### DIFF
--- a/packages/vue/src/composables/dialog.ts
+++ b/packages/vue/src/composables/dialog.ts
@@ -15,7 +15,7 @@ export function useDialog() {
 		/** Unmounts the dialog. Should be called after its closing animations. */
 		unmount: () => dialogStore.removeComponent(),
 		/** Whether the dialog is shown. */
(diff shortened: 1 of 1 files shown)


innocenzi:
Can you show me a snippet? I also think the `local` option might be too specific?

Tofandel:
It's just to allow this (as a lot of dialogs in framework have a v-model)

```vue
const { show } = useDialog();


<QDialog
    v-model="show"
>
  <QCard></QCard>
</QDialog>
```

Instead of something like
```vue
const { show, closeLocally } = useDialog();

<QDialog
    :model-value="show"
    @update:model-value="(v) => !v ? closeLocally() : null"
>
  <QCard></QCard>
</QDialog>
```

Because this particular dialog does not have a `@close` event

The idea is just to close it locally, instead of with a round trip as the background content is already loaded, so this is what makes most sense

innocenzi:
Makes sense, thank you!

PR #172, opened 2024-07-23
fix(table): correct `isPageSelected` logic

robin-dongbin:
fix(table): prevent 'isPageSelected' from returning true when no recordsEnsure that 'isPageSelected' computed property correctly handles the case
where there are no records in the table. It now returns false instead of
potentially evaluating to true due to the empty array every() call returning
true. This change aligns the behavior with the expected logic.

### DIFF
--- a/packages/vue/src/composables/table.ts
+++ b/packages/vue/src/composables/table.ts
@@ -166,7 +166,7 @@ export function useTable<
 		/** Deselects records on the current page. */
 		deselectPage: () => bulk.deselect(...table.value.records.map((record: RecordType) => getRecordKey(record))),
 		/** Whether all records on the current page are selected. */
-		isPageSelected: computed(() => table.value.records.every((record: RecordType) => bulk.selected(getRecordKey(record)))),
+		isPageSelected: computed(() => table.value.records.length > 0 && table.value.records.every((record: RecordType) => bulk.selected(getRecordKey(record)))),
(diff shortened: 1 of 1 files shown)


packages/vue/src/composables/table.ts
```diff
@@ -166,7 +166,7 @@ export function useTable<
 		/** Deselects records on the current page. */
 		deselectPage: () => bulk.deselect(...table.value.records.map((record: RecordType) => getRecordKey(record))),
 		/** Whether all records on the current page are selected. */
-		isPageSelected: computed(() => table.value.records.every((record: RecordType) => bulk.selected(getRecordKey(record)))),
+		isPageSelected: computed(() => table.value.records.length > 0 && table.value.records.every((record: RecordType) => bulk.selected(getRecordKey(record)))),
```
Tofandel, line 169:
```suggestion
		isPageSelected: computed(() => ! table.value.records.some((record: RecordType) => ! bulk.selected(getRecordKey(record)))),
```

Tofandel: Seems that you want to return false if you find any that you don't find selected instead

robin-dongbin, line 169: Yes, but I think the changes will make reading more difficult for me.

innocenzi:
Thanks!

PR #173, opened 2024-07-23
docs: correct viewmodel name in example

tppope:
change name of the public property $chirps to $comments base on the example below


innocenzi:
Thanks!

PR #174, opened 2024-07-23
Improvement for Preloading on touch devices

michaelsieminski:
When using the preloading feature with links, the default behavior is that the site is being preloaded on hover. On a touch device this is not possible so with this PR it will preload on mount if preloading is enabled in hover mode.

This is slightly opinionated as a lot of things in hybridly but I think it would improve the developer experience as it's a pretty default behavior you would want to have when you use preloading, to not break your site on mobile.

### DIFF
--- a/packages/vue/src/components/link.ts
+++ b/packages/vue/src/components/link.ts
@@ -56,7 +56,7 @@ export const RouterLink = defineComponent({
 					return
 				}
 
-				if (type === 'mount' && preloads !== 'mount') {
+				if (type === 'mount' && preloads !== 'mount' && !('ontouchstart' in window || navigator.maxTouchPoints > 0)) {
 					return
 				}
 


Tofandel:
Preloading is just an optimisation, I think preloading on mount is not a good idea because it will try to preload every link on the page, which defeats preloading purposes as well as uses a ton of Internet data which is not always desirable on mobile

michaelsieminski:
> Preloading is just an optimisation, I think preloading on mount is not a good idea because it will try to preload every link on the page, which defeats preloading purposes as well as uses a ton of Internet data which is not always desirable on mobile

I see, so you think preloading should just be disabled completly on Touch devices?

RobertBoes:
I think https://instant.page/ is a good inspiration for this. On desktop it would preload when the user hovers over a link, on mobile it would start preloading when the users touches the screen (which isn't a click yet), this would still save quite a few milliseconds

innocenzi:
Yeah I don't think this is the right fix for your issue. I just think preloading should be disabled when hover events can't be listened for, or by implementing what Robert suggests.

Let's close this and continue the discussion on Discord.

PR #175, opened 2024-07-24
fix(core): replace errors instead of merging them on response

owenconti:
I'm not sure about the error bag handling, not too familiar with error bags.

### DIFF
--- a/packages/core/src/router/router.ts
+++ b/packages/core/src/router/router.ts
@@ -197,7 +197,17 @@ export async function performHybridNavigation(options: HybridRequestOptions): Pr
 		// because the back-end sent back only the required properties.
 		if (payload.view && (options.only?.length ?? options.except?.length) && payload.view.component === context.view.component) {
 			debug.router(`Merging ${options.only ? '"only"' : '"except"'} properties.`, payload.view.properties)
-			payload.view.properties = merge(context.view.properties, payload.view.properties)
+			const mergedPayloadProperties = merge(context.view.properties, payload.view.properties)
+
+			// Overwrite errors with the errors coming in from the response instead of deeply merging them
+			// which prevents errors from being removed when they are not present in the response.
+			if (options.errorBag) {
+				(mergedPayloadProperties.errors as any)[options.errorBag] = (payload.view.properties.errors as any)[options.errorBag] ?? {}
+			} else {
+				mergedPayloadProperties.errors = payload.view.properties.errors
+			}
+
+			payload.view.properties = mergedPayloadProperties
 			debug.router('Merged properties:', payload.view.properties)
 		}
 


innocenzi:
Error bag handling works properly 👍

Tofandel:
I think this broke reactivity on `form.errors` I'll have to repro but since 0.7.13 using form.errors in a computed never triggers a second time

Edit: There is a reactivity issue on form.errors but not from this PR

innocenzi:
@Tofandel between 0.7.12 and 0.7.13, the only change that could have affected `form.errors` is this PR. Are you able to identify a version of Hybridly where this ever worked?

innocenzi:
I can't reproduce either, both `watch` and `computed` seem to work for me:

```ts
computed(() => login.errors)
watch(login.errors, (errors) => ...)
```

Feel free to open an issue or hop on Discord

Tofandel:
I'll be debugging this weekend but I think it might be only the nested errors with this issue, but no I tried with older versions and it didn't seem to work either, I looked at the code, everything looks ok as well so it might be a vue issue

Basically I just have

```js
const error = computed(() => form.errors.profile?.first_name)
```

Which never updates error when submitted with an error

But this works

```js
const error = computed(() => form.hasErrors ? form.errors.profile.first_name : null)
```

innocenzi:
That might be due to the way we update `errors`? https://github.com/hybridly/hybridly/blob/0.x/packages/vue/src/composables/form.ts#L239-L251

Tofandel:
That's what I checked and `setValueAtPath` seems okay, (Btw since there is @clickbar/dot-diver, maybe we could use `setByPath` instead since this function already exists there)

Since it does this basically

```js
errors.value.profile = {};
errors.value.profile.first_name = 'Some error';
```

So it should in theory trigger the errors ref but doesn't (I checked and using triggerRef(errors) after `setErrors` also fixes the issue)

Tofandel:
I found the culprit, it's `getByPath` of dot-diver, I'm guessing because of the hasOwnProperty check

https://play.vuejs.org/#eNp9VU1v20YQ/StjXkghLOUmPRSJlMIpckgPtdEavZhGwFBDaW1yl9hdSjYU/fe+2aVkxjaii5bz8ea9mVlyn1z0fbEdOHmfLFxtVe/JsR/6j6VWXW+spz1Zrmqvtpzj1ORUm64fPK/oQI01HaVIT0td6mbQiDNaEP6r2oEv/FXlN5n5dpdTj1NOWzHPaI9wP5/Tl4b8hoOPVoadTj3gta+UpgoWn4tf046prjTdDS7QC0kBqhAg1VB2JhiF0nU7rNhlaZHOxjKlR/0bcd/SMmZFs4VOC9qlPwj7QOgSyHanHLSipmao9IZc36pYNDBVWmy87lh7R5Ve0a5q70F3pyVohEJVrpFlmWnQXrWC2ChEC5BV642nvq1qDhVeiEIbgtixyjKULgKToE5iWmRpdpjFZSy2lKriaYylTNwKtvMP+FucsIqW9RoyfqFf4Xjz5tSnWPKeH5FzDL5Rt9H5rM708QYp0tuXtu/faS/NDR3+AeLmVOAVWreTSYXZRGZsrbHSCqxhtj+gBdEOsV2wxjXNIIfGYOTPPggCGnytOjaDz7IZLT+KZnq+qDGnCJVzSntrGtVy0Sjr/FdddZzCeg0VqYAecnp7fn4e8Sfbv2b/6THgSYm4BrkcwxUodeg3yZ6Du9gurK2k55k8zH6cMQLjnj5FFpZXQ81ZVg/WonPxbl1V1j8JI5JLEU9E/rFn09AYT2fLJaWRV4oRHaOO7iXcemjbiessDq1AR7wRtGJTucudvrKmZ+sfi7pq21f4xPxRsPxGLYNeMa4Cr6I9zHjiHoHCpRUcrCCC8rGXQH2+FG/RveNraZyvLEUxDnQc5B+TSZ6WJ4S8TH8a4gTo9Z2YhTkt5vH1iRcnHjx3uNue8US0WKltOBBpgFXtRV2zc+9pv4/sD4fFNztGnAqf3AeoBcg8oizmE+wkT7yDjkatiztnNN7iodVlInLA0172spOuTAAXC5QJZmV2fwWbt9j0o73ecH3/iv3OPYitTK4sO7ZbLpOTz1cWjKP7879/8wPOJ2dnVkOL6J84/2Fn2kE4xrBPWA3QnsQFtl/Ct0jp9bX7/OBZu6MoIRq3I+rGt+jPn0h/ovuu+C3kYZnQxa9btoKJBsJRvPs9OfwPJ2xysQ==

https://github.com/clickbar/dot-diver/pull/35

PR #185, opened 2024-10-13
chore: composer locks update

eshimischi:
Sorry missed updating of lock files in my previous PR, i think it's important too


eshimischi:
I have active PHP 8.3 and suppose it went too far with some sub-dependencies, it bet you'll update lock files later yourself

PR #186, opened 2024-10-14
fix(components): add `scope` variable to `HasScope` concern

tomykho:
Fix deprecation warnings for:
```
Creation of dynamic property Hybridly\\Refining\\Refine::$scope is deprecated in /vendor/hybridly/laravel/src/Components/Concerns/HasScope.php on line 9
```


innocenzi:
Thanks!

PR #187, opened 2024-10-21
fix(types): properly infer `useTable#paginator` type

tomykho:
This PR fixes `paginator` always return `never`.


innocenzi:
That seems wrong. Can you show me why is this change needed for you?

tomykho:
`paginator` returned from `useTable` is always `never`, It seems [this line](https://github.com/hybridly/hybridly/blob/0.x/packages/vue/src/composables/table.ts#L92C29-L92C115) always trigger `else` because of `AsRecordType<T>` instead of `T`:
```
Props[PropsKey] extends Table<RecordType, infer PaginatorKind> ? PaginatorKind : never
```
image

innocenzi:
Thanks—I'll have to check this out before merging, unsure why I haven't stumbled upon it yet

innocenzi:
Thank you!

PR #189, opened 2024-10-24
Support Laravel Folio

jamesj2:
This PR adds support for [Laravel Folio](https://laravel.com/docs/folio) for use with the Hybridly's [route](https://hybridly.dev/api/utils/route.html#route) helper.  Most of this is copied from [Ziggy pull #733](https://github.com/tighten/ziggy/pull/733).

Laravel Folio registers its routes very differently from how Laravel usually handles routes, so unfortunately this is a bit messy. We're more or less doing what Folio's [ListCommand](https://github.com/laravel/folio/blob/master/src/Console/ListCommand.php) is doing, modified for our needs.

Notes:
    We only deal with named Folio routes.


innocenzi:
I'm not familiar with Folio, does it not register routes in the router?

I'm also curious how you use Folio along with Hybridly?

Tofandel:
> does it not register routes in the router?

Looking at the doc, it seems to register routes on the fly by looking through directories with blade.php in them

I also don't see well how you could integrate this with hybridly which uses controllers to return hybrid response

A blade.php file seems like it would completely bypass all that, as such the routes should be properly handled as external routes and do a full reload on navigation

jamesj2:
> > does it not register routes in the router?
> 
> Looking at the doc, it seems to register routes on the fly by looking through directories with blade.php in them
> 
> I also don't see well how you could integrate this with hybridly which uses controllers to return hybrid response
> 
> A blade.php file seems like it would completely bypass all that, as such the routes should be properly handled as external routes and do a full reload on navigation

Folio is a page based router similar to vue router.  It looks through directories to build the routes and they are cachable.

I have a large app with many pages and it allows me to put the GET routing, middleware, controller and blade in the same file.  I originally started using it when I converted my app to [Splade](https://splade.dev/).  So I have a good amount of folio based routes already.

Here is a simple example of how I started using it with hybridly.  I return the hybridly response and don't have any blade content.

```php
// view/pages/index.blade.php
<?php
use function Laravel\Folio\{middleware, name, render};

middleware(['auth']);
name('home');

render(function () {
    // ...
    return hybridly('index', []);
});
```

innocenzi:
So, you are using that instead of controllers, if I understand correctly?

jamesj2:
@innocenzi Yes that's correct.

Tofandel:
I see, I could see this being a nice feature if you could put the blade file right next to the Vue file, so your code is not split across and hard to browse, but I also think this is something that would be best served by an extension of hybridly

I'm sure there are other routing strategies out there and just adding a configurator with hooks or some kind of similar solution, so another package can modify the routes returned by hybridly would be nice

@jamesj2 does it entirely replace your laravel router? Or do you have both laravel routes and folio routes?

jamesj2:
@Tofandel Yes I'm currently using both.  I'm still working on refactoring my app, plus folio only supports GET routes.  I love the idea of being able to providing middleware, routing, controllers and vue in the same file just like single file components.

innocenzi:
@jamesj2 I've been thinking about this, and I don't think we should implement Folio support directly, as this is too specific. We should allow the process to be overridden or hooked into, as suggested by @Tofandel

innocenzi:
Can you try https://github.com/hybridly/hybridly/pull/190?

PR #192, opened 2024-11-04
fix(refining)!: change `beingsWithStict` to `beginsWithStrict`

jamesj2:
Filters typo fix: beingsWithStrict should be beginsWithStrict


innocenzi:
Oops thank you!

PR #194, opened 2024-12-11
fix(routing): generate query arguments with optional parameters

Tofandel:
Given any route with an optional parameter

Eg php
```php
Route::get('test/{page?}')->name('test')
```

The `route` util would remove query args

Eg query arg `foo` is missing in the following
```js
route('test', {foo: 'bar', 'baz': 'baz'}) // /test?baz=baz
// expected /test?foo=bar&baz=baz
```

This is what currently happens
```
const missing = ['categories'];
missing.splice(missing.indexOf('page'), 1)
console.log(missing) // []
```


innocenzi:
Thanks!

PR #198, opened 2025-03-14
fix(vue): make `useQueryParameters` update after every navigation

### LINKED ISSUE #197: useQueryParameters is not reactive to router.local()
### Describe the bug

When changing query parameters via `router.local` useQueryParameters does not react to the url change

### Reproduction

See below

### Steps to reproduce

```ts

const useDrawerQuery = (drawerRef: Ref<boolean>, queryName: string) => {
  const params = useQueryParameters<{drawer: string}>();
  if (params.drawer === queryName) {
    drawerRef.value = true;
  }
  let ignore = false;
  console.log(params);
  watch(() => params.drawer, () => {
    console.log(params); // This is not called
    drawerRef.value = params.drawer === queryName;
    ignore = true;
  });
  watch(drawerRef, async (v) => {
    if (ignore) {
      ignore = false;
      return;
    }
    const url = new URL(window.location.toString());

    url.searchParams.delete('drawer', queryName);
    if (v) {
      url.searchParams.append('drawer', queryName);
    }
    await router.local(url.toString(), { preserveScroll: true, preserveState: true });
  });
};
const showDrawer = ref(false);
useDrawerQuery(showDrawer, 'test');

setTimeout(() => {
  showDrawer.value = true;
}, 2000):
```

### System information

Latest hybridly

### Used package manager

npm

### Logs

_No response_

### Validations

- [x] Read the [docs](https://hybridly.dev).
- [x] Check that there isn't [already an issue](https://github.com/hybridly/hybridly/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a Hybridly issue and not an issue related to something else (Vite, Vue...). For example, if it's a Vue SFC related bug, it should likely be reported to [vuejs/core](https://github.com/vuejs/core) instead.
- [x] Check that this is a concrete bug. For Q&A, join our [Discord server](https://discord.gg/uZ8eC7kRFV).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Additionally it would be nice if we could set query parameters in `useQueryParams` as well and use `router.local` like in the example to update the url

Tofandel:
Fixes #197 

The `after` hook only fires on a real request while the `navigated` hook fires for any hybrid navigation

### DIFF
--- a/packages/vue/src/composables/query-parameters.ts
+++ b/packages/vue/src/composables/query-parameters.ts
@@ -24,7 +24,7 @@ export function useQueryParameters<T extends Record<string, any> = Record<string
 	}
 
 	updateState()
-	registerHook('after', updateState)
+	registerHook('navigated', updateState)
 
 	return state as T
 }


innocenzi:
Thanks!

PR #199, opened 2025-03-14
feat(vite): allow the vite project to reside in a different folder than laravel

Tofandel:
Resubmitting #124


innocenzi:
I'm not against this, but I'm curious how you'd use it. Can you show me an example [tree](https://tree.nathanfriend.io/)?

Tofandel:
Sure, my projects usually look like this, but other kinds of structure could be used as well to make it reside almost anywhere by tweaking those configs
https://github.com/hybridly/hybridly/blob/78f856d65eeec7d0d28460aa76231fcfd571b240/packages/laravel/config/hybridly.php#L41-L44

```
└── project/
    ├── .hybridly/
    │   └── ...
    ├── app/
    │   └── ...
    ├── bootstrap/
    │   └── ...
    ├── config/
    │   └── ...
    ├── database/
    │   └── ...
    ├── resources/ #contained npm project
    │   ├── application/
    │   │   ├── main.ts
    │   │   └── root.blade.php
    │   ├── layouts/
    │   │   └── default.vue
    │   ├── views/
    │   │   └── index.vue
    │   ├── package.json
    │   ├── tailwind.config.ts
    │   ├── ... // Tons of other files
    │   └── vite.config.ts #  hybridly({...}, { laravelPath: path.resolve('..') })
    ├── routes/
    │   └── web.php
    ├── .env
    ├── artisan
    ├── composer.json
    └── package.json # optional: workspaces referencing resources/package.json, this allows for multiple packages in subfolders and avoids clutering the root with config files
```

The next step could be to allow for multiple hybridly workspaces, for instance if you wanted to build two separate npm projects for frontend and backend, but right now the limiting factor is the laravel app which only allows for one config and one root_directory

innocenzi:
Okay, I see what you want to do. To me it's a bit overkill, but I don't see a reason not to implement this. I'll have to test it before merging though

innocenzi:
Sorry it took so long to review. I tried the PR but I don't really like the new passing around of paths and the fact that the plugin has two parameters now, as well as the fact that `loadConfiguration` mutates the result of the artisan command. 

I might revisit when I add support for Tempest, but for now, I'm closing this, sorry

PR #201, opened 2025-03-20
fix: router.local() loses all existing properties

Tofandel:
By default since router.local stays on the component it makes little sense to not keep existing properties

### DIFF
--- a/packages/core/src/router/router.ts
+++ b/packages/core/src/router/router.ts
@@ -519,7 +519,7 @@ export async function performLocalNavigation(targetUrl: UrlResolvable, options?:
 			url,
 			view: {
 				component: options?.component ?? context.view.component,
-				properties: options?.properties ?? {},
+				properties: options?.properties ?? context.view.properties,
 				deferred: [],
 			},
 		},


innocenzi:
What was your use case that prompted this change? I'm pretty sure this was on purpose

Tofandel:
To add query parameters in the url and history state, so that a drawer or pop-up element on a page can have it's open close state persisted in the url and be closed by using the back browser button (good mobile UX).

Using router.local without any option just destroys the component it's on because all of the required props it's expecting become missing

Tofandel:
If its intentional I don't see why, because a component without any prop is going to have a hard time to do anything useful. And if you do need to remove them it's easy to be explicit by doing properties: {}

It's weird that the component, if not specified would stay the same but not the props

packages/core/src/router/router.ts
```diff
@@ -519,7 +519,7 @@ export async function performLocalNavigation(targetUrl: UrlResolvable, options?:
 			url,
 			view: {
 				component: options?.component ?? context.view.component,
-				properties: options?.properties ?? {},
+				properties: options?.properties ?? context.view.properties,
```
Tofandel, line 522: Maybe checking if options.component === context.view.component and no explicit properties then we keep the same properties and remove them otherwise

